PR #8, opened 2023-02-17
[DSERV-94] Adding CAQtl adapter

pedrohr:
Creating CAQtl (edge) and Open Chromatic Region (nodes)


data/adapters/encode_caqtl_adapter.py
```diff
@@ -0,0 +1,68 @@
+import csv
+
+from adapters import Adapter
+from adapters.helpers import build_variant_id, build_open_chromatic_region_id
+
+# Example Encode caQTL input file:
```
mingjiecn, line 6: Is this the file example from encode? I checked file in encode and the format seems to be different. Example here: https://www.encodeproject.org/annotations/ENCSR266TOD/

pedrohr: it is, we decided to use the archived file, not the star file. The default file is the result of processing of the archived/deprecated file, and it turns out we need the "raw" file with all columns

mingjiecn (approved):

PR #10, opened 2023-03-03
DSERV-100 automatic endpoints generated from config file

pedrohr:


data/schema-config.yaml
```diff
@@ -1,17 +1,13 @@
...
 topld:
   represented_as: edge
   inherit_properties: true
   label_in_input: topld
   label_as_edge: VARIANT_CORRELATION
   db_collection_name: variant_correlations
   db_collection_per_chromosome: false
+  relationship:
+    from: sequence variant
+    to: sequence variant
```
pedrohr, line 10: This is not the final list of endpoints in `schema-config.yml`. They should be defined later as needed per client applications.

docs/adding_endpoints.md
```diff
@@ -0,0 +1,86 @@
...
+    rsid: str
+```
+
+The `relationship` object can defined in any `edge` object and it must contain two keys: `from` and `to`.
+
+The values of `from` and `to` must be a singular class name of a node also defined in the `schema-config.yaml` file.
+
+In our example above, we are defining a relationship (sequence variant) -> (open chromatin region). This will automatically define the following endpoints, if the objects have an `accessible_via` object in their schema definitions:
+
+* `GET /variants/children`
```
mingjiecn, line 85: Can a node have more than one type of parent and more than one type of child? If so can we filter by type when we query the parent or child?

pedrohr: yes, it can have as many as defined by the `relationship` blocks. The endpoint will return all of them.

pedrohr: if they have children a `/children` endpoint will be created, similarly with parents

mingjiecn: I mean if a node have two types of children, can we filter the children by different relationship (edge)? something like /children?edge= topld

pedrohr: oh actually... I just realized the endpoint is missing the `id` field... I will update it

oh got it! it's a great suggestion, so far we don't support it, if we have a use case, we can add it, but as of now the endpoint just allows a broad exploration of the data

mingjiecn (approved):

docs/adding_endpoints.md
hitz: yeah we should add this for ontologies specifically as they have very different meanings, the edges.   There are also properties on edges that could could be filtered but no need to implement that without full use case.

PR #16, opened 2023-03-22
DSERV-111-schema-bug

mingjiecn:


hitz (commented):
this looks good to me, but @pedrohr should also review I think

pedrohr (approved):
💯 
looks great! thanks!

PR #18, opened 2023-03-27
DSERV-114-pre-commit

mingjiecn:
pre-commit setting is here: https://github.com/IGVF-DACC/igvf-catalog/blob/DSERV-114-pre-commit/.pre-commit-config.yaml. Let me know if this is good enough. Thank you!


pedrohr:
all python changes are ok

the eslint I have for my typescript is conflicting with these changes, is the circle-ci configured to use the same target as the repo:

```
"env": {
    "browser": true,
    "es2021": true
  }
```

and the compiler options from `tsconfig.json`?

mingjiecn:
close

PR #20, opened 2023-03-29
[DSERV-91] FAVOR adaptor

pedrohr:
Adding FAVOR adaptor


data/adapters/favor_adapter.py
```diff
@@ -0,0 +1,113 @@
...
+
+    DATASET = 'favor'
+
+    def __init__(self, filepath=None):
+        self.filepath = filepath
+        self.dataset = Favor.DATASET
+
+        super(Favor, self).__init__()
+
+    def parse_info_metadata(self, info):
```
mingjiecn, line 63: Can you comment function parse_info_metadata on how metadata is processed here? Thank you!

mingjiecn (changes requested):
after comment function parse_info_metadata on how metadata is processed, it should be good to go.

mingjiecn (approved):

PR #22, opened 2023-04-05
[DSERV-120] Adding json output for Favor adaptor

pedrohr:
Favor has an object (the frequencies) as a property. 

If we want to access it on Arango queries, it must be stored as an object, which is not possible via CSV import.

Since Biocypher only supports CSV outputs, I wrote an independent JSON output, all the ontology functionalities continue to work, just the writer of the file that was changed.


hitz (approved):
Should we have some place to document how IDs are hashed for each entity?  This would be useful when querying  outside API maybe?

PR #23, opened 2023-04-07
Dserv-109 asb adapter

shengchd:
Adapter for loading allele-specific binding events from ADASTRA


data/adapters/adastra_asb_adapter.py
```diff
@@ -0,0 +1,95 @@
...
+            TF_uniprot_csv = csv.reader(TF_uniprot_mapfile, delimiter='\t')
+            next(TF_uniprot_csv)
+            for row in TF_uniprot_csv:
+                if row[0] == TF_name:
+                    return row[1]  # return uniprot id of the TF (i.e. protein)
+        return None
+
+    def get_cell_ontology_id(self):
+        cell_name = self.filepath.split('/')[-1].split('@')[1].split('.')[0]
+        ontology_priority_list = ['CL:', 'UBERON:', 'EFO:']
```
pedrohr, line 35: we should make this a class constant on line 14

data/adapters/adastra_asb_adapter.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, filepath, TF_ids, cell_ontologies):
+        self.filepath = filepath
+        self.TF_ids = TF_ids
+        self.cell_ontologies = cell_ontologies
+        self.dataset = ASB.DATASET
+
+        super(ASB, self).__init__()
+
+    def get_TF_uniprot_id(self):
+        TF_name = self.filepath.split('/')[-1].split('@')[0]
```
pedrohr, line 24: in python, variables that start with a capital letter are interpreted as constants, we should rename it to `tf_name` for example (even though it all works fine :P)

pedrohr (approved):

PR #24, opened 2023-04-13
DSERV-122-load-gencode

mingjiecn:


config/development.json
```diff
@@ -6,11 +6,11 @@
     "port": 2023
   },
   "database": {
-    "connectionUri": "http://arangodb:8529",
+    "connectionUri": "http://127.0.0.1:8529",
```
pedrohr, line 9: this will break the docker container

Dockerfile
```diff
@@ -8,10 +8,10 @@ RUN wget -q https://download.arangodb.com/arangodb310/DEBIAN/Release.key -O- | a
...
 
-COPY . /app
 WORKDIR /app
-
-RUN pip3 install -r data/requirements.txt && \
-  npm install
+COPY data/requirements.txt .
+RUN pip3 install -r requirements.txt
+COPY . /app
+RUN npm install
```
pedrohr, line 15: why do we need this change? the second COPY will replace the first COPY

config/development.json
mingjiecn: I will create a production json config then.

PR #25, opened 2023-04-13
Situate ontology classes in biolink schema

slobentanzer:
I added inheritance of the ontologies you add to the graph (clo, mondo, etc) to be subclasses of the Biolink `ontology class`. If we don't define them, they will be detached in the ontological hierarchy. As is, BioCypher will complain about that.

What is your purpose in adding the ontologies into the graph in this verbose manner?


pedrohr:
💯 
thank you!

PR #30, opened 2023-04-18
DSERV-123 Modifying ccre adapter according to schema adjustments

pedrohr:


data/dev_setup.py
```diff
@@ -38,7 +38,6 @@
     'go_edges': Ontology(ontology='GO', type='edge'),
     'efo_nodes': Ontology(ontology='EFO', type='node'),
     'efo_edges': Ontology(ontology='EFO', type='edge'),
-    'ccre': CCRE(filepath='./samples/ccre_example.bed.gz'),
     'UniProtKB_protein': Uniprot(filepath='./samples/uniprot_sprot_human_sample.dat.gz'),
```
pedrohr, line 41: duplicated w/ line 26

data/dev_setup.py
```diff
@@ -38,7 +38,6 @@
     'go_edges': Ontology(ontology='GO', type='edge'),
     'efo_nodes': Ontology(ontology='EFO', type='node'),
     'efo_edges': Ontology(ontology='EFO', type='edge'),
-    'ccre': CCRE(filepath='./samples/ccre_example.bed.gz'),
```
pedrohr, line 41: duplicated: line 26

data/schema-config.yaml
```diff
@@ -314,6 +314,7 @@ regulatory region:
     start: int
     end: int
     biochemical_activity: str
+    biochemical_activity_description: str
```
mingjiecn, line 317: Since we add one more property **biochemical_activity_description**, should we also need to add it in filter_by and return in accessible_via?

mingjiecn (approved):

data/schema-config.yaml
pedrohr: just added as a return. This is a text field, it would be a good fit for a `filter_by` exact string match.

PR #39, opened 2023-04-25
DSERV-142 Adding source fields to CCREs

pedrohr:


data/schema-config.yaml
```diff
@@ -329,6 +329,9 @@ ccre regulatory region:
     end: int
     biochemical_activity: str
     biochemical_activity_description: str
+    type: str
```
mingjiecn, line 332: I think we also need to return those added properties in accessible_via

mingjiecn (changes requested):
need to return those added properties in accessible_via

PR #42, opened 2023-04-26
DSERV-147 Upgrading biocypher

pedrohr:


data/schema-config.yaml
```diff
@@ -247,6 +247,7 @@ transcribed from:
     - SIO:010081
 
 caqtl:
+  is_a: ontology class
```
pedrohr, line 250: The new version of Biocypher forces you to declare inheritance if your node is not present in the BiolinkML ontology.

If the class is new and doesn't have a direct relationship to any existing class in Biolink, we can add a generic `ontology class` and `ontology relationship`.

mingjiecn (approved):

PR #46, opened 2023-05-10
DSERV-204 TRPC Ontology endpoints

pedrohr:


src/client.ts
```diff
@@ -8,13 +8,47 @@ async function main (): Promise<void> {
...
+    term: 'liver'
+  })
+
+  console.log(terms)
+
+  terms = await trpc.ontologyGoTermBP.query({
+    term_name: 'synthesis'
+  })
+
+  console.log(terms)
```
hitz, line 39: should all these console.log() be in here?

PR #47, opened 2023-05-12
DSERV-96-enhancer-gene-link

mingjiecn:


data/adapters/encode_enhancer_gene_adapter.py
```diff
@@ -0,0 +1,113 @@
...
+                    continue
+                chr = row[0]
+                start = row[1]
+                end = row[2]
+                class_name = row[4]
+                regulatory_element_id = build_regulatory_region_id(
+                    class_name, chr, start, end)
+                score = row[9]
+
+                if self.label == 'element_gene_link':
```
hitz, line 88: I think we should remove _link both here and from the variant_gene_link.  I know I wrote it that way though.  I haven't written that ticket though so I guess leave it be for now.

data/adapters/encode_enhancer_gene_adapter.py
```diff
@@ -0,0 +1,113 @@
...
+                    gene_id = row[6]
+
+                    _id = regulatory_element_id + '_' + gene_id
+                    _source = 'regulatory_regions/' + regulatory_element_id
+                    _target = 'genes/' + gene_id
+                    _props = {
+                        'score': score,
+                        'source': self.source,
```
hitz, line 96: this needs chr and start and stop too?

data/schema-config.yaml
```diff
@@ -865,3 +865,67 @@ gtex variant to gene expression association:
...
+regulatory region:
+  represented_as: node
+  label_in_input: regulatory_region
+  db_collection_name: regulatory_regions
+  db_collection_per_chromosome: false
+  db_indexes:
+    coordinates:
+      type: zkd
+      fields: start:long,end:long
+  accessible_via:
```
hitz, line 905: this whole stanza needs to be copied to the other schema

hitz (changes requested):

data/schema-config.yaml
hitz: WRONG

data/adapters/encode_enhancer_gene_adapter.py
hitz: never mind this is fine

hitz (approved):

pedrohr (approved):

PR #48, opened 2023-05-15
DSERV-155 Ontology adaptor improvements

pedrohr:


mingjiecn (changes requested):
The old ontology schemas need to be dropped? Other than that it looks good

pedrohr:
> The old ontology schemas need to be dropped? Other than that it looks good

Just talked to Ben, we can remove the old ontology collections and schemas. We are dropping the `type` field, so we are ready to merge.

mingjiecn (approved):

PR #62, opened 2023-06-12
DSERV-167 load HOCOMOCO motifs

shengchd:
- Adapter for loading TF motifs from HOCOMOCO pwm files
- Creating nodes of motifs and edges between motifs and proteins


pedrohr (approved):
💯

data/adapters/motif_adapter.py
```diff
@@ -0,0 +1,127 @@
...
+                             ','.join(Motif.ALLOWED_LABELS))
+
+        self.filepath = filepath
+        self.label = label
+        self.dataset = label
+        if label == 'motif':
+            self.type = 'nodes'
+            self.collection = 'motifs'
+        else:
+            self.type = 'edges'
```
pedrohr, line 45: it should be `edge` here

and `node` for line 42

pedrohr (changes requested):
`self.type` should be singular

pedrohr (approved):

PR #63, opened 2023-06-13
DSERV-191-biocypher-update

mingjiecn:


data/tests/test_arango_db.py
```diff
@@ -170,7 +170,7 @@ def test_arangodb_generates_import_statements_for_nodes(mock_client):
         db = ArangoDB()
 
         cmds = db.generate_import_statement(
-            'header-test.csv', ['data1.csv', 'data2.csv'], 'test_collection', 'node')
+            'header-test.csv', ['data1.csv', 'data2.csv'], 'test_collection', 'node', True)
```
pedrohr, line 174: we need an extra test with False just to guarantee the `--translate` parameter won't be passed

pedrohr (changes requested):
We just need a test with `has_edge_id` as False.

data/db/arango_db.py
```diff
@@ -79,7 +70,7 @@ def generate_import_statement(self, header_path, data_filenames, collection, ele
 
             if element_type == 'edge':
                 cmd += '--create-collection-type edge --translate ":START_ID=_from" --translate ":END_ID=_to"'
-                if self.has_edge_key(header_path):
+                if has_edge_id:
                     cmd += ' --translate "id=_key"'
```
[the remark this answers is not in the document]
mingjiecn: Yes in Biocypher the default setting is True.

data/tests/test_arango_db.py
pedrohr: thank you 🙏

pedrohr (approved):
👍

PR #64, opened 2023-06-13
DSERV-189 Adding db-xref relationships

pedrohr:


data/adapters/ontologies_adapter.py
```diff
@@ -134,17 +136,42 @@ def process_edges(self, predicate):
...
+
+                        # only accepting IDs in the form <ontology>:<ontology_id>
+                        if len(str(to_node).split(':')) != 2:
+                            print(
+                                'Unsupported format for xref: ' + str(to_node))
+                            continue
+
+                        to_node_key = str(to_node).replace(':', '_')
+
+                        if from_node_key == to_node_key:
```
mingjiecn, line 157: is `if str(to_node) == str(from_node):` duplicated with `if from_node_key == to_node_key:`?

mingjiecn (changes requested):
Looks like the key definition for ontology related collections is changed? Can you please also update the doc: https://github.com/IGVF-DACC/igvf-catalog/blob/main/docs/collection_key_definition.md. Thanks!

data/adapters/ontologies_adapter.py
pedrohr: it looks duplicated, but once you replace ':' by '_' you can still have duplicates. These are edge cases.

PR #65, opened 2023-06-14
DSERV-192 Merging regulatory regions and ccre collections

pedrohr:


data/adapters/encode_enhancer_gene_adapter.py
```diff
@@ -105,9 +105,17 @@ def process_file(self):
                         'chr': chr,
                         'start': start,
                         'end': end,
-                        'class': class_name,
+                        'type': 'candidate_cis_regulatory_element',
                         'source': self.source,
                         'source_url': self.source_url
                     }
 
+                    if class_name != 'enhancer':
```
mingjiecn, line 113: should be `if class_name == 'enhancer':`

mingjiecn (changes requested):
Also accessible_via and properties need to be updated in regulatory region schema.

data/adapters/encode_enhancer_gene_adapter.py
pedrohr: oh good catch!

mingjiecn (approved):

mingjiecn:
One more thing. Can you update the collection key definition here since we merge two schemas? https://github.com/IGVF-DACC/igvf-catalog/blob/main/docs/collection_key_definition.md
Thanks!

PR #69, opened 2023-06-16
DSERV-196-gencode-id

mingjiecn:


data/dev_setup.py
```diff
@@ -57,13 +57,14 @@
...
 args = parser.parse_args()
-
 dry_run = args.dry_run
 create_indexes = args.create_indexes
 adapters = args.adapter or ADAPTERS.keys()
 
 if not dry_run:
     ArangoDB().setup_dev()
+    ADAPTERS['gencode_genes'] = GencodeGene(
+        filepath='./samples/gencode_sample.gtf', gene_alias_file_path='./samples/Homo_sapiens.gene_info.gz', dry_run=False)
```
pedrohr, line 68: this looks misplaced

pedrohr (changes requested):
just one line on dev_setup.py, everything else is great

data/dev_setup.py
mingjiecn: dry_run is set to True in GencodeGene adapter by default. I need to pass the dry_run in GencodeGene adapter if dry_run is False.

pedrohr: oh, I see! There are other adapters in the same situation. The dev_setup script today doesn't support dry_run=True for adapters that skip biocypher. Adding this support should be a separate ticket I think. I can work on that. We can leave dry_run default to False in the adapter for now.

mingjiecn: Sure

pedrohr (approved):

PR #70, opened 2023-06-19
DSERV-198 updating db collection names

pedrohr:


data/schema-config.yaml
```diff
@@ -497,7 +497,7 @@ gtex splice variant to gene association:
   represented_as: edge
   label_in_input: GTEx_splice_QTL
   label_as_edge: GTEX_SPLICE_QTL
-  db_collection_name: variant_gene_links
+  db_collection_name: variant_genes
```
mingjiecn, line 500: should be variants_genes?

data/schema-config.yaml
```diff
@@ -535,7 +535,7 @@ gtex variant to gene expression association:
   represented_as: edge
   label_in_input: GTEx_eqtl
   label_as_edge: GTEx_eQTL
-  db_collection_name: variant_gene_links
+  db_collection_name: variant_genes
```
mingjiecn, line 538: should be variants_genes?

mingjiecn (changes requested):
See my comment for requested changes. Thank you!

data/schema-config.yaml
pedrohr: yes, good catch! just pushed the change

mingjiecn (approved):

mingjiecn (approved):

PR #71, opened 2023-06-22
DSERV-184-gene-alias

mingjiecn:


data/adapters/gencode_gene_adapter.py
```diff
@@ -96,6 +100,11 @@ def process_file(self):
                     split_line[GencodeGene.INDEX['info']:])
                 gene_id = info['gene_id']
                 id = gene_id.split('.')[0]
+                alias = alias_dict.get(id, None)
```
pedrohr, line 103: not a big deal, but the default behavior of `.get` when it doesn't find a key is to return None, so no need to make it explicit

pedrohr (approved):

data/adapters/gencode_gene_adapter.py
mingjiecn: Will change it then merge. Thank you so much!

PR #73, opened 2023-06-27
DSERV-163-gene-gene

mingjiecn:


data/adapters/coxpresdb_adapter.py
```diff
@@ -0,0 +1,42 @@
...
+        self.file_path = file_path
+        self.dataset = 'coxpresdb'
+        self.label = 'coxpresdb'
+        self.source = 'CoXPresdb'
+        self.source_url = 'https://coxpresdb.jp/'
+
+        super(Coxpresdb, self).__init__()
+
+    def process_file(self):
+        with open('./samples/entrez_ensembl.pkl', 'rb') as f:
```
pedrohr, line 24: I think we should move this to a different folder since it's required for data loading and not a sample for dev testing.

Also just a quick comment on how this pkl was generated, or in the future if we have to modify we might not remember.

pedrohr (changes requested):
Minor changes: just creating a new folder for supporting files and including one comment.

pedrohr (approved):

PR #74, opened 2023-06-28
DSERV-176 GWAS Studies Adapter

pedrohr:


data/adapters/gwas_adapter.py
```diff
@@ -0,0 +1,263 @@
...
+                '_key': key,
+                '_from': variant_id,
+                '_to': ontology_term_id,
+                'genes': genes.get(row[0]),
+                'equivalent_ontology_term': equivalent_term_id,
+                'study_id': row[3],
+                'lead_chrom': row[4],
+                'lead_pos': row[5],
+                'lead_ref': row[6],
+                'lead_alt': row[7],
```
hitz, line 127: There is an interesting denormalization here; these values should always match the _from field.  Not sure what to do with this information, but it's true.

data/schema-config.yaml
```diff
@@ -663,6 +663,65 @@ gaf:
...
+    direction: str
+    beta: str
+    beta_ci_lower: str
+    beta_ci_upper: str
+    odds_ratio: str
+    oddsr_ci_lower: str
+    oddsr_ci_upper: str
+    pval_mantissa: str
+    pval_exponent: str
+    pval: str
```
hitz, line 693: these are numbers but I think maybe schema doesn't support floats?

hitz (approved):

docs/collection_key_definition.md
```diff
@@ -16,6 +16,7 @@ The schema for each collecion we load into our database is defined in schema-con
 | gaf                                               | edge | go_gene_product_links               | Y | {annotation_dict}  |
 | gtex splice variant to gene association           | edge | variant_gene_links                  | N | {variant_id}\_{gene_id}\_{ontology_term_id} |
 | gtex variant to gene expression association       | edge | variant_gene_links                  | N | {variant_id}\_{gene_id}\_{ontology_term_id} |
+| variant to phenotypic feature association       | edge | variants_phenotypes                  | Y | {variant_id}\_{ontology_term_id} |
```
mingjiecn, line 19: collection key definition doesn't match since three collections are added. Need to add key definition for studies, studies_variants, and studies_variants_phenotypes.

mingjiecn (changes requested):
collection key definition doesn't match since three collections are added. Need to add key definition for studies, studies_variants, and studies_variants_phenotypes.

mingjiecn (approved):

PR #76, opened 2023-07-05
DSERV-205 TRPC Regulatory Regions Endpoints

pedrohr:


src/routers/datatypeRouters/_helpers.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function preProcessRegionParam (input: paramsFormatType): paramsFormatType {
+  const newInput = input
+  if (input.region !== undefined) {
+    const regex: RegExp = /(chr\w):(\d*)-(\d*)/
+
+    if (regex.test(input.region as string)) {
+      const breakdown = regex.exec(input.region as string)
+      if (breakdown === null || breakdown.length < 4) {
+        throw new TRPCError({
```
mingjiecn, line 13: Can refactor the code so only throw TRPCError once?

mingjiecn (changes requested):
see my comment for requested change. Thank you!

src/routers/datatypeRouters/_helpers.ts
pedrohr: done

mingjiecn (approved):

PR #79, opened 2023-07-12
DSERV-120 Adding special parser for chr x and y

pedrohr:


mingjiecn (approved):
One thing to mention is that the key format in annotations is different. The annotations key for chrx and chry is camelcase. The annotations key for other chrs is lowercase. But the content for annotation is so different so unified key format will not help anyway.

PR #80, opened 2023-07-14
DSERV-206 Variants API

pedrohr:


src/routers/datatypeRouters/nodes/variants.ts
```diff
@@ -0,0 +1,113 @@
+import { z } from 'zod'
+import { publicProcedure } from '../../../trpc'
+import { loadSchemaConfig } from '../../genericRouters/genericRouters'
+import { RouterFilterBy } from '../../genericRouters/routerFilterBy'
+import { RouterFilterByID } from '../../genericRouters/routerFilterByID'
+import { preProcessRegionParam, paramsFormatType } from '../_helpers'
+
+const schema = loadSchemaConfig()
+
+const frequencySources = z.enum([
```
pedrohr, line 10: this is a sample of sources, I will recalculated all possible sources once all data is loaded in the database

data/schema-config.yaml
```diff
@@ -41,19 +41,22 @@ sequence variant:
         - rsid
   accessible_via:
     name: variants
-    description: 'Retrieve variants data. Example: chr = chr1'
-    filter_by: _id, rsid
-    return: _id, chr, pos, rsid, ref, alt, qual, filter, format
+    description: 'Retrieve variant data. For example: region = chr1:1157520-1158189 or funseq_description = noncoding or rsid = rs2045642915'
```
mingjiecn, line 44: is `funseq_description` one of the properties? I don't see it in properties.

src/routers/__tests__/datatypeRouters/variants.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    test('accepts pagination', () => {
+      let inputParsing = router._def.inputs[0].parse({ funseq_description: 'noncoding' })
+      expect(inputParsing.page).toEqual(0)
+
+      inputParsing = router._def.inputs[0].parse({ funseq_description: 'noncoding', page: 1 })
+      expect(inputParsing.page).toEqual(1)
+    })
+
+    test('accepts regulatory region query format', () => {
```
mingjiecn, line 29: Need to change to variant query instead of regulatory region query.

src/routers/datatypeRouters/_helpers.ts
```diff
@@ -16,16 +16,21 @@ function validRegion (region: string): string[] | null {
   return null
 }
 
-export function preProcessRegionParam (input: paramsFormatType): paramsFormatType {
+export function preProcessRegionParam (input: paramsFormatType, field: string | undefined = undefined): paramsFormatType {
```
mingjiecn, line 19: what is this `field` parameter for?

src/routers/genericRouters/routerFilterBy.ts
```diff
@@ -55,7 +55,7 @@ export class RouterFilterBy implements Router {
       }
 
       if (queryParams[element] !== undefined) {
-        if (this.filterByRange.includes(element)) {
+        if (this.filterByRange.includes(element) || element.startsWith('annotations.freq')) {
```
mingjiecn, line 58: Should it be && not ||?

mingjiecn (changes requested):
See my comment for request changes. Thank you!

src/routers/genericRouters/routerFilterBy.ts
pedrohr: this is an ||, we can't have dot notation in the schema file yet. This is a special case that our schema doesn't support.

data/schema-config.yaml
pedrohr: yes, it's part of the `annotations` object. Not all variants have this but we still need to allow search

src/routers/datatypeRouters/_helpers.ts
pedrohr: it's the field which a region query will be applied to, I just renamed the variable just in case

mingjiecn (approved):

PR #81, opened 2023-07-19
DSERV-214 Genes API

pedrohr:


mingjiecn (approved):

mingjiecn (changes requested):
Forgot one thing, We also need to query gene by item in alias. I don't see it in code?

mingjiecn (approved):

src/routers/datatypeRouters/nodes/genes.ts
```diff
@@ -0,0 +1,109 @@
...
+  source_url: z.any(),
+  alias: z.array(z.string()).optional().nullable()
+})
+
+const schemaObj = schema.gene
+const router = new RouterFilterBy(schemaObj)
+const routerID = new RouterFilterByID(schemaObj)
+const routerFuzzy = new RouterFuzzy(schemaObj)
+
+async function conditionalSearch (input: paramsFormatType): Promise<any[]> {
```
pedrohr, line 80: I'm possibly refactoring this method into a helper in the future if I find a way to dynamically remove keys from objects in typescript

PR #84, opened 2023-07-20
DSERV-133-pathway

mingjiecn:


data/adapters/reactome_pathway_adapter.py
```diff
@@ -0,0 +1,111 @@
...
+                    to_json = {
+                        '_key': id,
+                        'name': name,
+                        'species': species,
+                        'source': 'Reactome',
+                        'source_url': 'https://reactome.org/'
+                    }
+
+                    try:
+                        query = 'https://reactome.org/ContentService/data/query/' + id
```
pedrohr, line 59: I'm thinking of ways of avoiding this, so we don't depend on their API to be functional... a few options I thought of:

* do they offer any data download that we can keep somewhere?
* run this script once and save all responses in a file and keep that versioned

do you have any other ideas?

pedrohr (approved):
Dataset is small! All good for the API solution 👍

data/schema-config.yaml
```diff
@@ -776,3 +776,109 @@ gene to gene coexpression association:
...
+    range: pathway
+    represented_as: edge
+    label_in_input: parent_pathway_of
+    label_as_edge: PARENT_PATHWAY_OF
+    db_collection_name: pathways_pathways
+    db_collection_per_chromosome: false
+    relationship:
+      from: pathway
+      to: pathway
+    properties:
```
hitz, line 864: add type: parent

data/schema-config.yaml
```diff
@@ -776,3 +776,109 @@ gene to gene coexpression association:
...
+    range: pathway
+    represented_as: edge
+    label_in_input: child_pathway_of
+    label_as_edge: CHILD_PATHWAY_OF
+    db_collection_name: pathways_pathways
+    db_collection_per_chromosome: false
+    relationship:
+      from: pathway
+      to: pathway
+    properties:
```
hitz, line 882: add type: child

hitz (changes requested):

hitz (approved):

PR #93, opened 2023-08-08
DSERV-230 Autocomplete endpoint

pedrohr:


data/schema-config.yaml
```diff
@@ -388,12 +388,12 @@ protein:
...
     text:
       type: inverted
-      fields: dbxrefs[*]
+      fields: name
   accessible_via:
     name: proteins
     description: 'Retrieve protein data. Example: name = 1433B_HUMAN'
     filter_by: _id, name
-    fuzzy_text_search: dbxrefs[*]
+    fuzzy_text_search: name
```
mingjiecn, line 396: Why remove dbxrefs from indexing and fuzzy search? I thought we also want to search for dbxrefs. It will be slow if no index?

mingjiecn (approved):

data/schema-config.yaml
pedrohr: the search results didn't end up well, if you search for `CAD` for example, you would ids like `CTD:12345`. It wasn't great. So Meenakshi and I decided to change to an exact match for now and have only protein names to be in the fuzzy search indexes.

PR #96, opened 2023-08-11
DSERV-231 Variant <-> Variant API

pedrohr:


data/schema-config.yaml
```diff
@@ -274,6 +280,10 @@ transcribed from:
   relationship:
     from: transcript
     to: gene
+  accessible_via:
+    name: genes_transcripts
+    description: 'Retrieve edge data between the relationship of genes and transcripts.'
```
mingjiecn, line 285: Do we need to add filter here for query?

pedrohr: oh we do, it's coming in the next PR with the variants_genes endpoints with all specific details. For this endpoint, we just need a simple declaration so it won't break.

mingjiecn (approved):

PR #99, opened 2023-08-23
DSERV-178 load orphanet disease-gene

shengchd:
Adapter for loading disease-gene edges from Orphanet xml file


pedrohr (approved):
💯

pedrohr:
I fixed the pep8 issue on main, you can ignore that alert.

PR #100, opened 2023-08-26
DSERV-164 load DepMap gene-term links

shengchd:
Adapter for loading gene-term edges from DepMap collection


data/dev_setup.py
```diff
@@ -67,6 +68,7 @@
     'cellosaurus_relationships': Cellosaurus('./samples/cellosaurus_example.obo.txt', type='edge'),
     'drug': PharmGKB('./samples/pharmGKB', type='node'),
     'variant_drug': PharmGKB('./samples/pharmGKB', type='edge'),
+    'gene_term': DepMap('./samples/DepMap/CRISPRGeneDependency_transposed_example.csv'),
```
mingjiecn, line 71: Should keep it consistant to pass type here

mingjiecn (changes requested):
Please see my comment. Thank you!

data/adapters/depmap_adapter.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class DepMap(Adapter):
+    SOURCE = 'DepMap'
+    SOURCE_URL = 'https://depmap.org/portal/'
+    GENE_ID_MAPPING_PATH = './data_loading_support_files/DepMap_gene_id_mapping.tsv'
+    CELL_ONTOLOGY_ID_MAPPING_PATH = './data_loading_support_files/DepMap_model.csv'
+
+    CUTOFF = 0.5  # only load genes with dependency scores greater or equal to 0.5 for each cell
+    SKIP_BIOCYPHER = True
```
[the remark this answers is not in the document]
shengchd: I have gene_dependency as float values, and I thought biocyper couldn't handle that.

data/adapters/depmap_adapter.py
```diff
@@ -0,0 +1,117 @@
...
+
+        with open(self.filepath, 'r') as depmap_file:
+            # read model ids on the first row
+            model_ids = next(depmap_file).strip().split(',')[1:]
+            # record the column index of each model ID (start from 2nd column)
+            model_ids_column_mapping = {}
+            for column_index, model_id in enumerate(model_ids):
+                model_ids_column_mapping[column_index] = model_id
+                # check CVCL id mapping for all models once first
+                cell_ontology_id = self.cell_ontology_id_mapping[model_id]['cell_ontology_id']
```
mingjiecn, line 55: This check seems to be redundant.  You also checked it again later at line 73.

mingjiecn (changes requested):
Apart from the possible duplication, the code looks good. One thing to make it better: for the mapping files, you can save them as pickle files then you can load and use them directly, you don't have to make a dict before using them in this way. Also you can include the script for making gene symbol to gene id mapping in the scripts folder so we have a reference how the mapping is generated.

data/adapters/depmap_adapter.py
shengchd: Yes it's a bit redundant. I wanted to check them and output all cells missing ids at once first, before iterating over all genes (and getting missing ids message for every gene in that cell), if that makes sense.

shengchd:
> Apart from the possible duplication, the code looks good. One thing to make it better: for the mapping files, you can save them as pickle files then you can load and use them directly, you don't have to make a dict before using them in this way. Also you can include the script for making gene symbol to gene id mapping in the scripts folder so we have a reference how the mapping is generated.

Thanks for the suggestion! I'll make the gene mapping file as pkl then. The model file was downloaded directly from their portal, so I won't change that. Will also include the mapping script.

PR #105, opened 2023-09-12
DSERV-248 Genes to Genes Endpoints

pedrohr:


src/routers/genericRouters/routerEdges.ts
```diff
@@ -82,6 +82,31 @@ export class RouterEdges extends RouterFilterBy {
...
+    if (filters) {
+      filters = ` AND ${filters}`
+    }
+
+    const query = `
+        FOR record IN ${this.edgeCollection}
+          FILTER (record._from == '${id}' OR record._to == '${id}') ${filters}
+          ${this.sortByStatement(sortBy)}
+          LIMIT ${page * QUERY_LIMIT}, ${QUERY_LIMIT}
+          RETURN {[record._from == '${id}' ? '${this.sourceSchemaName}' : '${this.targetSchemaName}']: UNSET(DOCUMENT(record._from == '${id}' ? record._to : record._from), '_rev', '_id'), ${this.dbReturnStatements}}
```
mingjiecn, line 104: Can you add an example for return here? It is hard to image what the return looks like. Thank you!

mingjiecn (changes requested):
Can you add a example return for function getCompleteBidirectionalByID? Thank you!

mingjiecn (approved):

src/routers/genericRouters/routerEdges.ts
mingjiecn: Thanks a lot!

PR #108, opened 2023-09-23
DSERV-236 GWAS Endpoints

pedrohr:


data/schema-config.yaml
```diff
@@ -1034,6 +1034,88 @@ child pathway of:
...
+    lead_ref: str
+    lead_alt: str
+    direction: str
+    beta: str
+    beta_ci_lower: str
+    beta_ci_upper: str
+    odds_ratio: str
+    oddsr_ci_lower: str
+    oddsr_ci_upper: str
+    pval_mantissa: str
```
mingjiecn, line 1093: the properties provided in schema don't match the properties used in adapter and test. For example  `pval_mantissa` here but `p_val_mantissa` in adapter, `pval` here but `p_val` in adapter and test.

mingjiecn (changes requested):

data/schema-config.yaml
pedrohr: good catch, just pushed the fix. This never failed because we are skipping Biocypher.

mingjiecn (approved):

PR #109, opened 2023-09-26
DSERV-268 Reading ENV variables for configuration

pedrohr:


src/env.ts
```diff
@@ -22,6 +24,18 @@ let config = envConfig
...
+    config.environment = 'production'
+    config.host.protocol = process.env.IGVF_CATALOG_PROTOCOL ?? config.host.protocol
+    config.host.hostname = process.env.IGVF_CATALOG_HOSTNAME ?? config.host.hostname
+    config.host.port = parseInt(process.env.IGVF_CATALOG_PORT ?? DEFAULT_PORT) ?? config.host.port
+    config.database.connectionUri = process.env.IGVF_CATALOG_ARANGODB_URI ?? config.database.connectionUri
+    config.database.dbName = process.env.IGVF_CATALOG_ARANGODB_DBNAME ?? config.database.dbName
+    config.database.auth.username = process.env.IGVF_CATALOG_ARANGODB_USERNAME ?? config.database.auth.username
+    config.database.auth.password = process.env.IGVF_CATALOG_ARANGODB_PASSWORD ?? config.database.auth.password
+  }
 }
```
ottojolanki, line 39: This looks like a good way to mix and match if I understand correctly? When running locally using a remote database we can pass the secrets using `docker compose -e IGVF_CATALOG_ARANGODB_PASSWORD ...` and still read most of the stuff from the file right?

ottojolanki:
Reading from the environment seems good. I do not understand how to run local catalog instance using a database that runs on a remote machine using docker-compose.

ottojolanki (changes requested):
I think some more work is required to be able to configure the arangoDB instance that we use?

src/env.ts
pedrohr: Yes, those are all available config variables. We can keep most of them in a config file and change them anytime we need by passing the secrets.

ottojolanki (approved):

PR #118, opened 2023-10-10
DSERV-212-spdi

mingjiecn:


data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+# check instruction here: https://github.com/biocommons/biocommons.seqrepo
+# for each file, an output file will be genereated.
+# each row in output file cotains: id in catalog, chr, pos, ref, alt, spdi, hgvs
+
+
+def main():
+    dp = create_dataproxy('seqrepo+file:///usr/local/share/seqrepo/2018-11-26')
+    translator = Translator(data_proxy=dp)
+    input_folder = '/home/ubuntu/datasets/favor'
+    ouput_folder = '/home/ubuntu/datasets/favor_output'
```
pedrohr, line 63: this only works on a node instance, should we make this configurable so we can run it locally?

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+            file_chr = file_name_split[1]
+        else:
+            file_chr = file_name_split[0]
+        output_name = f'{file_chr}_spdi.tsv'
+        output_path = os.path.join(ouput_folder, output_name)
+        file_path = os.path.join(input_folder, file)
+        with open(output_path, 'w', newline='') as csvfile:
+            writer = csv.writer(csvfile, delimiter='\t')
+
+            with gzip.open(file_path, 'rt') as input_file:
```
pedrohr, line 77: this will crash for FAVOR, usually unzipping the larger chromosomes takes a long time (> 15 minutes sometimes), we might have to ingest the decompressed files here

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+                        pos,
+                        ref,
+                        alt
+                    )
+                    # Only use translator if the ref and alt is more than one base.
+                    if len(ref) == 1 and len(alt) == 1:
+                        chr_ref = chr_map[chr]
+                        pos_spdi = int(pos) - 1
+                        # example SPDI: NC_000024.10:10004:C:G
+                        spdi = f'{chr_ref}:{pos_spdi}:{ref}:{alt}'
```
pedrohr, line 103: should we refactor this into a `build_spdi_id` method?

data/scripts/variants_spdi.py
pedrohr: another refactoring tip is that all hard coded strings should be constants

pedrohr (changes requested):
I added a few refactoring tips, and I'm concerned about extracting FAVOR on the fly, it takes a long time. chr2 when unzipped is almost half a terabyte.

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+                reader = csv.reader(input_file, delimiter='\t')
+                row = next(reader)
+                while not (row[0].startswith('#Chrom') or row[0].startswith('#CHROM')):
+                    row = next(reader)
+                for row in reader:
+                    chr = row[0]
+                    pos = row[1]
+                    ref = row[3]
+                    alt = row[4]
+                    if pos == 'NA' or ref == 'NA' or alt == 'NA':
```
shengchd, line 90: you probably won't see NA in those three fields in vcf files, I think they are usually labeled as '.' if it's unavailable.

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+    '13': 'NC_000013.11',
+    '14': 'NC_000014.9',
+    '15': 'NC_000015.10',
+    '16': 'NC_000016.10',
+    '17': 'NC_000017.11',
+    '18': 'NC_000018.10',
+    '19': 'NC_000019.10',
+    '20': 'NC_000020.11',
+    '21': 'NC_000021.9',
+    '22': 'NC_000022.11',
```
shengchd, line 30: I downloaded one of the vcf file, seems they have a bad naming for chromosome 22: ##contig=<ID=22>. You might want to adjust that key mapping here.

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,124 @@
...
+                        continue
+                    if chr not in chr_map.keys():
+                        continue
+                    id = build_variant_id(
+                        chr,
+                        pos,
+                        ref,
+                        alt
+                    )
+                    # Only use translator if the ref and alt is more than one base.
```
shengchd, line 100: I think you mean ref 'or' alt, the code below is correct.

data/scripts/variants_spdi.py
mingjiecn: You can run it locally. I have comment in the script regarding the instruction to set it up locally. I will use args to pass in the input and output file

data/scripts/variants_spdi.py
mingjiecn: sure

data/scripts/variants_spdi.py
mingjiecn: I copied this over from favor adapter so I just assume maybe there is some NA field?

data/scripts/variants_spdi.py
mingjiecn: Will fix the comment

data/scripts/variants_spdi.py
mingjiecn: Since translate to spdi and hgvs both require translator to generate allele first when ref or alt is more than one base long, which is one of the steps that takes time, it is hard to separate the generate SPDI and HGVS into two different methods. I will just use one function to generate SPDI and HGVS so I just need to generate allele once if needed.

shengchd (approved):

pedrohr (approved):
💯

mingjiecn:
Previously I also use vrs translator for generate HGVS, which is extremely slow probably due to the hgvs package dependency. I talked to Ben and he suggest that not use the translator for HGVS generation. This way the speed is much faster.  Also use vrs translator can generate the HGVS for the special case duplication and since using insertion for duplication is also valid, I will not check if the insertion is a duplication or not in the code.

data/scripts/variants_spdi.py
```diff
@@ -46,26 +47,20 @@ def build_allele(chr, pos, ref, alt, translator):
...
         chr_ref = CHR_MAP[chr]
         pos_spdi = int(pos) - 1
         # example SPDI: NC_000024.10:10004:C:G
         spdi = f'{chr_ref}:{pos_spdi}:{ref}:{alt}'
-
-        pos_hgvs = f'g.{pos}'
-        # example HGVS: NC_000024.10:g.10005C>G
-        hgvs = f'{chr_ref}:{pos_hgvs}{ref}>{alt}'
     else:
         allele = build_allele(chr, pos, ref, alt, translator)
```
shengchd, line 58: It might be possible to also return `valid_ref_seq` from `_from_gnomad`, which should just be `del_seq` you are getting below.

data/scripts/variants_spdi.py
```diff
@@ -78,13 +73,38 @@ def convert_spdi(spdi, seq):
...
+        else:
+            hgvs = f'{chr_ref}:g.{pos_hgvs_start}_{pos_hgvs_end}del'
+    # it is a insertion. We will not generate specail case hgvs - duplication, since using insertion for dup is valid.
+    # you can check if the hgvs is valid here: https://api.ncbi.nlm.nih.gov/variation/v0/#/HGVS/get_hgvs__hgvs__contextuals
+    else:
+        pos_hgvs_start = spdi_pos + len(del_seq)
+        pos_hgvs_end = spdi_pos + len(del_seq) + 1
+        insert_seq_hgvs = ins_seq[len(del_seq):]
+        hgvs = f'{chr_ref}:g.{pos_hgvs_start}_{pos_hgvs_end}ins{insert_seq_hgvs}'
+    return hgvs
```
shengchd, line 98: We will also need to catch `delins` type, I think the main logic will be to check if `del_seq` is a subset of `ins_seq` or in the other way around.

shengchd (changes requested):
It looks good except we need to catch `delins` type also. And seems _to_hgvs is using a different normalization method than _to_spdi (3' shifting vs expanded mode from VOCA), which is dependent on hgvs package and can be slow.

data/scripts/variants_spdi.py
```diff
@@ -0,0 +1,159 @@
...
+def build_hgvs_from_spdi(spdi):
+    ins_seq = spdi.split(':')[-1]
+    del_seq = spdi.split(':')[2]
+    spdi_pos = int(spdi.split(':')[1])
+    chr_ref = spdi.split(':')[0]
+    if len(ins_seq) == 1 and len(del_seq) == 1:
+        hgvs = f'{chr_ref}:g.{spdi_pos + 1}{del_seq}>{ins_seq}'
+    # check if this variant is a deletion
+    elif len(del_seq) > len(ins_seq):
+        pos_hgvs_start = spdi_pos + 1 + len(ins_seq)
```
shengchd, line 103: I missed this last time, shouldn't pos_hgvs_start always be `spdi_pos +1 `? (ie. the starting position of ref/ins seq, in 1-based coordinate)

shengchd (commented):
I'm not sure if `pos_hgvs_start` is correct.

shengchd (approved):
Looks good! I was wrong on the `pos_hgvs_start` calculation, it is correct.

PR #119, opened 2023-10-12
DSERV-288 Running arangodb on systemd

pedrohr:


config/development.json
```diff
@@ -12,5 +12,6 @@
         "username": "igvf",
         "password": "igvf"
     }
-  }
+  },
+  "cluster": ["0.0.0.0"]
```
pedrohr, line 16: just an example of the format. In production, this will have all IPs of the nodes of the cluster

devops/generate_systemd.py
```diff
@@ -0,0 +1,105 @@
...
+if __name__ == '__main__':
+    # Get the number of input arguments.
+    num_args = len(sys.argv)
+
+    if num_args < 1:
+        print('Node number required (1, 2, 3, or 4). Usage: $ python3 generate_systemd.py <node_number>')
+        exit(1)
+
+    try:
+        node_number = int(sys.argv[1])
```
mingjiecn, line 92: should also check the value of the node number to make sure it is in [1, 2, 3, 4].

mingjiecn (changes requested):
Looks good! See my comment for one small request

devops/generate_systemd.py
pedrohr: good catch! just added

mingjiecn (approved):

PR #121, opened 2023-10-19
DSERV-278 API Acceptance Tests

pedrohr:


tests/acceptance_tests/api_test.py
```diff
@@ -0,0 +1,102 @@
...
+        url = API + endpoint
+
+        for test in TESTS[endpoint]:
+            if test.get('params_url'):
+                url = API + endpoint.format(**test['params_url'])
+
+            response = requests.get(url, params=test.get('params'))
+
+            assert response.status_code == 200
+            print(response.json())
```
mingjiecn, line 99: should remove this print statement?

mingjiecn (approved):

tests/acceptance_tests/api_test.py
pedrohr: this only gets printed if the test fails, and it's very useful to debug the response

PR #124, opened 2023-10-19
DSERV-264 add biosample hyperedge for variants-genes

shengchd:
- create variants_genes_terms collection for GTEx adapters
- iterate over all tissues inside adapter


pedrohr (approved):
💯 
really cool it now iterates on the folder! no need to extra scripts

PR #126, opened 2023-10-26
DSERV-293 Complexes API Endpoints

pedrohr:


data/schema-config.yaml
```diff
@@ -1161,6 +1161,15 @@ complex:
   label_in_input: complex
   db_collection_name: complexes
   db_collection_per_chromosome: false
+  db_indexes:
```
mingjiecn, line 1164: should also add persistent index for complex_source and evidence_code

src/routers/genericRouters/routerFuzzy.ts
```diff
@@ -43,18 +43,21 @@ export class RouterFuzzy extends RouterFilterBy implements Router {
         LIMIT ${page * QUERY_LIMIT}, ${QUERY_LIMIT}
         RETURN { ${this.dbReturnStatements} }
     `
-
     const cursor = await db.query(query)
     return await cursor.all()
   }
 
-  async autocompleteSearch (term: string, page: number, apiReturn: boolean = false, customFilter: string = ''): Promise<any[]> {
+  async autocompleteSearch (term: string, page: number, apiReturn: boolean = false, customFilter: string = '', overWriteSearchField: string = ''): Promise<any[]> {
```
mingjiecn, line 50: Can you write the function documentation for function usage, parameters, returned value. It is hard for me to understand what is the meaning of some of the parameters. Thank you!

mingjiecn (changes requested):
Please check my comment for changes request. Thank you!

data/schema-config.yaml
pedrohr: they are not being used by any of the endpoints as search parameters

mingjiecn: Oh, I though filtering by complex_source and evidence_code will need those two fields to be indexed?

pedrohr: I think we only need to return them

src/routers/genericRouters/routerFuzzy.ts
pedrohr: added :)

mingjiecn (approved):

PR #129, opened 2023-10-27
DSERV-313-protein-name

mingjiecn:


data/schema-config.yaml
```diff
@@ -412,15 +412,16 @@ protein:
         - dbxrefs[*]
     text:
       type: inverted
-      fields: name
+      fields: name, rec_full_name
```
pedrohr, line 415: this is currently not supported, but my branch for Complexes has support for inverted fields with multiple fields, so we just need to run the script after merging that one first.

pedrohr (approved):

data/schema-config.yaml
mingjiecn: Sounds good. I will wait for the merge. Thanks!

PR #133, opened 2023-11-02
DSERV-313-protein-schema

mingjiecn:


pedrohr (approved):

src/routers/datatypeRouters/nodes/proteins.ts
```diff
@@ -10,13 +10,15 @@ const schema = loadSchemaConfig()
...
   name: z.string().optional(),
+  full_name: z.string().optional(),
   dbxrefs: z.string().optional(),
   page: z.number().default(0)
 })
 
 export const proteinFormat = z.object({
   _id: z.string(),
   name: z.string(),
+  full_name: z.string(),
```
pedrohr, line 21: do all proteins have a 'full_name' defined? just checking if this field should be optional or not

mingjiecn: emm, I will change it to optional

pedrohr (approved):

PR #144, opened 2023-11-16
DSERV-311 DSERV-208 DSERV-195 Ontology Data Clean Up

pedrohr:


data/schema-config.yaml
```diff
@@ -1015,7 +1015,7 @@ child pathway of:
 study:
   represented_as: node
   is_a: ontology class
-  label_in_input: gwas_study
```
pedrohr, line 1018: these were fixes preventing the dev_setup scrip from working

data/adapters/ontologies_adapter.py
```diff
@@ -58,62 +58,67 @@ class Ontology(Adapter):
...
+        if ontology not in Ontology.ONTOLOGIES.keys():
+            raise ValueError('Ontology not supported.')
+
+        self.dataset = 'ontology_term'
+
         self.dry_run = dry_run
-        if type == 'node':
-            self.dataset = 'ontology_term'
-        else:
-            self.dataset = 'ontology_relationship'
```
mingjiecn, line 67: how is the data loaded to ontology_relationship if don't specify `self.dataset = 'ontology_relationship'` when loading edge data?

mingjiecn (changes requested):
self.dataset = 'ontology_relationship' is deleted in the code. I don't understand how to load data to ontology_terms_ontology_terms collection?

data/adapters/ontologies_adapter.py
pedrohr: this is not used since we skip biocypher: https://github.com/IGVF-DACC/igvf-catalog/pull/144/files#diff-fc12bd16c5c066c3e11b1e1672cacb2f2adef2faece4507b95b28980fdbc123eL11

this is a requirement by the parent class `Adapter`, it could be any value there

mingjiecn: I think the parent class Adapter need `self.dataset` to get the correct schema config then get the collection name from the schema. Without specify `self.dataset` to `ontology_relationship`, you will get the wrong collection name - `ontology_terms` when loading edge data.

pedrohr: ooo I see the issue now at the `save_to_arango` method, good catch! I'm trying to have each ontology to process its own nodes and edges at once, it speeds up a lot, I will find a way to change it

mingjiecn (approved):

mingjiecn (approved):

PR #148, opened 2023-11-18
DSERV-309 Changing response format of ontology graph endpoints

pedrohr:


src/routers/genericRouters/routerEdges.ts
```diff
@@ -947,4 +947,29 @@ export class RouterEdges extends RouterFilterBy {
     const cursor = await db.query(query)
     return await cursor.all()
   }
+
+  // A --> B --> C, given B, return all A's if opt == parent, or return all C's if opt == children
+  // assuming A, B, C are all of the same type
+  async getChildrenParents (id: string, opt: string, sortBy: string, page: number): Promise<any[]> {
```
mingjiecn, line 953: param `id` doesn't seem to be used anywhere in this function. Can you please explain how to use it?

src/routers/genericRouters/routerTransitiveClosure.ts
```diff
@@ -38,7 +32,7 @@ export class RouterTransitiveClosure implements Router {
     this.nodeCollectioName = nodeCollectionName
   }
 
-  async getPaths (from: string, to: string): Promise<any> {
+  async getPaths (from: string, to: string, fields: string[]): Promise<any> {
```
mingjiecn, line 35: Can you also add comment for this function? I am not quite sure what does this function do. Thank you!

mingjiecn (dismissed):
Please check my comments. Thank you!

src/routers/genericRouters/routerEdges.ts
pedrohr: `id` is used on line 962: https://github.com/IGVF-DACC/igvf-catalog/pull/148/files#diff-51a1524cf522a929b1661fef4f94e784ac63775650ba678bc003809ec76530e1R962

PR #170, opened 2024-01-06
DSERV-371 Merging gene and protein search

pedrohr:


mingjiecn (approved):

mingjiecn (changes requested):
It is still hard for me to understand what those two functions can do. Can you please add a real application example in our catalog for each of those functions: secondaryEdgeCollection and getSelfAndTransversalSourceEdges. Thank you!

mingjiecn (approved):

PR #171, opened 2024-01-08
DSERV-354-gene

mingjiecn:
What I did :
1. Update the adapters to load mouse data: gene, transcript, protein and edges between them.
2. Fix bug in adapter to schema reading and for loading data in local arangodb database.


data/adapters/__init__.py
```diff
@@ -14,7 +14,7 @@ def __init__(self):
             schema_configs = yaml.safe_load(config)
 
             for c in schema_configs:
-                if schema_configs[c].get('label_in_input') == self.dataset:
+                if schema_configs[c].get('label_in_input') == self.label:
```
pedrohr, line 17: why do we need this change? many adapters don't set labels

data/adapters/gencode_adapter.py
```diff
@@ -60,41 +67,41 @@ def process_file(self):
...
-                if self.type == 'transcript':
+                if self.label in ['gencode_transcript', 'mm_gencode_transcript']:
                     props = {
                         'transcript_id': info['transcript_id'],
                         'transcript_name': info['transcript_name'],
                         'transcript_type': info['transcript_type'],
                         'chr': data[Gencode.INDEX['chr']],
                         # the gtf file format is [1-based,1-based], needs to convert to BED format [0-based,1-based]
-                        'start:long': str(int(data[Gencode.INDEX['coord_start']]) - 1),
-                        'end:long': data[Gencode.INDEX['coord_end']],
```
pedrohr, line 71: we need to keep `:long` for numeric values because of the old biocypher compatibility

pedrohr (changes requested):
* the :long suffix is required for numerical values
* discussion on `dataset` -> `label`

PR #178, opened 2024-01-27
DSERV-93-load-MPRA-regulatory-elements

shengchd:
Adapter for loading regulatory regions nodes & regulatory regions to biosamples edges from ENCODE MPRA experiments.


data/schema-config.yaml
```diff
@@ -934,6 +934,29 @@ regulatory element to gene expression association to biosample to donor:
...
+  label_in_input: regulatory_region_biosample
+  label_as_edge: REGULATORY_REGION_BIOSAMPLE
+  db_collection_name: regulatory_regions_biosamples
+  db_collection_per_chromosome: false
+  relationship:
+    from: regulatory region
+    to: ontology term
+  accessible_via:
+    name: regulatory_regions_biosamples
+    return: _id, source, source_url
```
mingjiecn, line 949: You defined a lot properties that are not returned here, for example element_name, strand and activity_score. I think we should return those properties.

mingjiecn (approved):
Looks good. But I think you forgot to return some of the properties for search result in schema?

data/schema-config.yaml
shengchd: Yes I will add them.

shengchd:
> Looks good. But I think you forgot to return some of the properties for search result in schema?

I will add them, thank you!

PR #180, opened 2024-01-31
DSERV-383 Go terms annotations endpoints

pedrohr:


src/routers/genericRouters/routerEdges.ts
```diff
@@ -1119,6 +1120,48 @@ export class RouterEdges extends RouterFilterBy {
     return await cursor.all()
   }
 
+  // Given id for A, and A --(edge)--> B, return B's and edge's
```
mingjiecn, line 1123: Can you please added a real application we use for function `getTargetAndEdgeSet` and `getSourceAndEdgeSet`?

mingjiecn (changes requested):
Can you add some detailed examples in function comments so I can understand the functions a little better? Thank you!

src/routers/datatypeRouters/edges/go_terms_proteins.ts
```diff
@@ -0,0 +1,93 @@
...
+async function goTermsSearch (input: paramsFormatType): Promise<any[]> {
+  const query = input.query as string
+  const page = input.page as number
+
+  let response
+
+  // assuming an ID will match either a gene or a protein
+  const genes = await geneIds(query)
+  if (genes.length !== 0) {
+    response = await genesProteinsGoRouter.getSourceAndEdgeSet(genes[0], page)
```
mingjiecn, line 56: If I understand it correctly, function `getSourceAndEdgeSet` is to get source go term for given target. The target here is protein. I am not sure how can you search gene in protein collection?

pedrohr: we still don't have the data for GO terms to genes, we are just leaving the support here

mingjiecn: I think maybe for given id for go term search, if it is protein id, get the go term using edge collection go_terms_proteins. If it is gene id, find corresponding protein id for that gene id first, then get the go term.

mingjiecn (changes requested):
I think for given id for go term search, if it is protein id, get the go term using edge collection go_terms_proteins. If it is gene id, find corresponding protein id for that gene id first, then get the go term.

src/routers/datatypeRouters/edges/go_terms_proteins.ts
pedrohr: I will adapt the code and do the gene -> proteins traversal and then fetch the go terms

mingjiecn (approved):

mingjiecn (approved):

mingjiecn (approved):

src/routers/datatypeRouters/edges/go_terms_annotations.ts
```diff
@@ -0,0 +1,150 @@
...
+  // assuming an ID will match either transcripts or proteins
+  const transcripts = await transcriptIds(query)
+  if (transcripts.length !== 0) {
+    annotations = transcripts
+  } else {
+    annotations = await proteinIds(query)
+  }
+
+  if (annotations.length > 0) {
+    const query = `
```
pedrohr, line 77: we've decided to start a big refactoring and start bringing the queries from the routerEdge file into each independent router so it's easier to read and maintain. We will keep the routerEdges file as is for now in the project in case it's needed in the future

mingjiecn (approved):

PR #188, opened 2024-03-01
DSERV-391-load-AFR-QTLs

shengchd:
Adapters for loading AFGR QTLs (splice QTL, eQTL, caQTL).


data/adapters/AFGR_caqtl_adapter.py
```diff
@@ -0,0 +1,79 @@
...
+                region_chr, region_pos_start, region_pos_end = row[8].split(
+                    ':')
+                regulatory_region_id = build_regulatory_region_id(
+                    region_chr, region_pos_start, region_pos_end, class_name=AFGRCAQtl.CLASS_NAME
+                )
+
+                if self.label == 'regulatory_region':
+                    _id = regulatory_region_id
+                    _props = {
+                        'chr': region_chr,
```
pedrohr, line 47: we need to add a `name` property to all nodes now... in this case, would `'name': _id` work?

data/adapters/AFGR_eqtl_adapter.py
```diff
@@ -0,0 +1,54 @@
...
+
+                variants_genes_id = hashlib.sha256(
+                    (variant_id + '_' + gene_id + '_' + AFGREQtl.SOURCE).encode()).hexdigest()
+
+                _id = variants_genes_id
+                _source = 'variants/' + variant_id
+                _target = 'genes/' + gene_id
+
+                _props = {
+                    'biological_context': 'ontology_terms/' + AFGREQtl.ONTOLOGY_TERM,
```
pedrohr, line 45: same need for a `name` property

pedrohr (changes requested):
code looks great! We just need to add a default `name` property to all nodes now for consistency of the API

data/adapters/AFGR_caqtl_adapter.py
shengchd: yes it works

pedrohr (approved):

PR #189, opened 2024-03-01
DSERV-364 Normalizing name properties of nodes

pedrohr:


src/routers/datatypeRouters/descriptions.ts
```diff
@@ -27,7 +27,7 @@ export const descriptions = {
     Set verbose = true to retrieve full info on the corresponding variants and genes.<br> Example: variant_id = 22f170e54c30a59e737beba20444f192201126f0b1415a7c9a106d1d01fe98d0, log10pvalue = gte:2',
   genes_variants: 'Retrieve variant-gene pairs from GTEx eQTLs & splice QTLs by Ensembl gene ids.<br> The following parameters can be used to set thresholds on -log10 p_value: gt (>), gte (>=), lt (<), lte (<=).<br> \
     Set verbose = true to retrieve full info on the corresponding variants and genes.<br> Example: gene_id = ENSG00000187642, log10pvalue = gte:2',
-  motifs: 'Retrieve transcription factor binding motifs from HOCOMOCO.<br> Example: name = STAT3_HUMAN',
+  motifs: 'Retrieve transcription factor binding motifs from HOCOMOCO.<br> Example: tf_name = STAT3_HUMAN',
```
pedrohr, line 30: `tf_name` is now `name` and `name` is now `_key`. We are adding more sources into the same collection with same tf_name, so name should be unique.

src/routers/datatypeRouters/edges/complexes_proteins.ts
```diff
@@ -11,7 +11,7 @@ const proteinComplexFormat = z.object({
   source: z.string().optional(),
   source_url: z.string().optional(),
   protein: z.string().or(z.array(proteinFormat)).optional(),
-  complex: z.string().or(z.array(complexFormat)).optional()
+  complex: z.string().or(complexFormat).optional()
```
pedrohr, line 14: cases like this would always return an array with a single value. I finally fixed it for many cases. There are still a few left but they are more complicated and will be easier when we start our refactoring and copying all queries from the routerEdge files into the routers.

src/routers/datatypeRouters/edges/diseases_genes.ts
```diff
@@ -90,9 +90,22 @@ async function conditionalGeneSearch (input: paramsFormatType): Promise<any[]> {
...
+  name: true
+}).merge(z.object({
+  gene_name: z.string().trim().optional(),
+  source: z.string().trim().optional(),
+  page: z.number().default(0),
+  verbose: z.enum(['true', 'false']).default('false')
+})).merge(associationTypes).transform(({gene_name, ...rest}) => ({
+  name: gene_name,
+  ...rest
+}))
```
pedrohr, line 104: the original `genesQueryFormat` has the updated `name` property from genes.
in any edge endpoint where we import the gene query object, we would have the `name` property, which is confusing in edge endpoints where more than one node is involved.
So here, we import the original query format, rename it to `gene_name` for the endpoint, and then rename the property back to `name`

mingjiecn (approved):
It is a huge update! It looks great to me.

PR #192, opened 2024-03-15
DSERV-404 Adding indexes

pedrohr:


data/schema-config.yaml
```diff
@@ -339,6 +339,15 @@ variant to regulatory region:
   label_as_edge: VARIANT_REGULATORY_REGION
   db_collection_name: variants_regulatory_regions
   db_collection_per_chromosome: false
+  db_indexes:
+    query:
+      type: persistent
+      fields:
+        - source
+        - label
```
pedrohr, line 347: these are not being used by any endpoints, internal queries only, so no need to combine fields

PR #210, opened 2024-04-04
DSERV-413 rename qtl effect_size

shengchd:
- Renamed `slope` in GTEx and `beta` in AFGR to `effect_size`
- Updated adapter and api scripts accordingly


src/routers/datatypeRouters/edges/variants_genes.ts
```diff
@@ -122,6 +110,48 @@ async function qtlSearch (input: paramsFormatType): Promise<any[]> {
...
+  `
+
+  const targetQuery = `FOR otherRecord IN genes
+  FILTER otherRecord._id == record._to
+  RETURN {${getDBReturnStatements(geneSchema).replaceAll('record', 'otherRecord')}}
+  `
+
+  const query = `
+    FOR record IN variants_genes
+    ${filterBy} ${customFilters}
```
pedrohr, line 142: is there a case where we have `customFilters` but no `filterBy`?

shengchd: I think `customFilters` are always added when there's filter on effect_size/log10pvalue, which are stored in `filterBy`, so shouldn't be such cases?

pedrohr (approved):

data/adapters/gtex_sqtl_adapter.py
```diff
@@ -103,12 +103,13 @@ def process_file(self):
                                     'chr': variant_id_ls[0],
                                     'biological_context': self.ontology_term_mapping.get(filename_biological_context) or biological_context,
                                     'sqrt_maf': to_float(line_ls[5]),
+                                    'p_value': pvalue,
                                     'log10pvalue': log_pvalue,
                                     'pval_nominal_threshold': to_float(line_ls[9]),
                                     'min_pval_nominal': to_float(line_ls[10]),
-                                    'slope': to_float(line_ls[7]),
+                                    'effect_size': to_float(line_ls[7]),
                                     'slope_se': to_float(line_ls[8]),
```
ottojolanki, line 111: Would it make sense to also change `slope_se` to `effect_size_se`?

shengchd: Yes that makes sense, I can update it together with the `source_url` change in another PR, since the reloading will take a while.

PR #221, opened 2024-04-16
DSERV-461-add-GTEx-filename-url

shengchd:
- use ontology term names from UBERON in `biological_context`
- use specific filename in `source_url`
- change `slope_se` -> `effect_size_se`
- add index for `biological_context` in variants_genes and variants_genes_terms
- modify `biological_context` for AFGR sQTLs and eQTLs


ottojolanki (approved):
LGTM

pedrohr (approved):

PR #232, opened 2024-04-24
DSERV-412-load-E2G-CRISPR

shengchd:
Adapter for loading E2G training data from CRISPR in `regulatory_regions` and `regulatory_regions_genes`.


data/adapters/encode_E2G_CRISPR_adapter.py
```diff
@@ -0,0 +1,100 @@
...
+
+        with open(self.filepath, 'r') as crispr_file:
+            crispr_csv = csv.reader(crispr_file, delimiter='\t')
+            next(crispr_csv)
+            for row in crispr_csv:
+                regulatory_region_coordinate = ','.join(row[:3])
+
+                significant = row[17]
+
+                if self.regulatory_region_nodes.get(regulatory_region_coordinate) is None:
```
pedrohr, line 96: this is good! usually people write this conditional using `in`, something like:
`if regulatory_region_coordinate not in self.regulatory_region_nodes:`

pedrohr (approved):

data/adapters/encode_E2G_CRISPR_adapter.py
shengchd: got it thanks!

PR #238, opened 2024-05-14
DSERV-463 Adding y2ave convert script

pedrohr:


data/scripts/variants_y2ave_convert.py
```diff
@@ -0,0 +1,69 @@
...
+
+        annotations_keys = ['funseq_description', 'varinfo',  'vid', 'variant_vcf',  'variant_annovar', 'start_position',  'end_position', 'ref_annovar', 'alt_annovar', 'ref_vcf', 'alt_vcf', 'bravo_af', 'bravo_an', 'af_total', 'af_asj_female', 'af_eas_female', 'af_afr_male', 'af_female', 'af_fin_male', 'af_oth_female',
+                            'af_ami', 'af_oth', 'af_male', 'af_ami_female', 'af_afr', 'af_eas_male', 'af_sas', 'af_nfe_female', 'af_asj_male', 'af_raw', 'af_oth_male', 'af_nfe_male', 'af_asj', 'af_amr_male', 'af_amr_female', 'af_sas_female', 'af_fin', 'af_afr_female', 'af_sas_male', 'af_amr', 'af_nfe', 'af_eas', 'af_ami_male', 'af_fin_female']
+
+        annotations = {}
+        for key in annotations_keys:
+            if key in orig_json['annotations']:
+                annotations[key] = orig_json['annotations'][key]
+
+        if list(annotations.keys()) != ['varinfo']:
```
mingjiecn, line 48: why remove this variant `if list(annotations.keys()) != ['varinfo']`?

data/scripts/variants_y2ave_convert.py
```diff
@@ -0,0 +1,69 @@
...
+                'source': 'FAVOR',
+                'source_url': 'http://favor.genohub.org/',
+                'y2ave': True
+            }
+
+            output_file = chrs_files[orig_json['chr']]
+            output_file.write(json.dumps(processed_json))
+            output_file.write('\n')
+
+for chrr in chrs_files.keys():
```
mingjiecn, line 68: typo here?

mingjiecn (changes requested):
Just have a question regarding filtering the data. A small typo need to be fixed.

data/scripts/variants_y2ave_convert.py
pedrohr: `chr` is a reserved word in python for character. I can still change the name

data/scripts/variants_y2ave_convert.py
pedrohr: there are many variants with empty annotations in the new files (when they only have the `varinfo` property). We are only looking for the annotations from the y2ave variants.

data/scripts/variants_y2ave_convert.py
mingjiecn: oh, haha, you are right!

mingjiecn (approved):

PR #241, opened 2024-05-16
DSERV-341 DSERV-474 Updating frequency filters 

pedrohr:


src/routers/datatypeRouters/_helpers.ts
```diff
@@ -134,7 +141,7 @@ export function getFilterStatements (
 
       // 'interesect' is a reserved parameter for intersectional region search
       // 'annotation.freq' is a special case for variant data
```
mingjiecn, line 143: Needs to update the comment here since annotation.freq is removed

mingjiecn (approved):
Just need to update the comment accordingly.

PR #246, opened 2024-05-22
DSERV-477-add-regulatory-region-type

shengchd:
- add type and source for E2G CRISPR data


pedrohr (approved):

src/routers/datatypeRouters/edges/regulatory_regions_genes.ts
```diff
@@ -25,6 +25,17 @@ const edgeSources = z.object({
   ]).optional()
 })
 
+const regulatoryRegionType = z.enum([
+  'candidate_cis_regulatory_element',
+  'enhancer',
+  'CRISPR_tested_element'
+])
+
+const biochemicalActivity = z.enum([
```
mingjiecn, line 34: biochemicalActivity for regulatory region is already defined in regulatory_regions.ts.

shengchd: yeah the issue is not all regulatory_regions have edges to genes, so I redefined biochemicalActivity enum list here, which is a subset from the one in regulatory_regions.ts.

mingjiecn (approved):

PR #255, opened 2024-06-06
DSERV-491 update variants phenotypes endpoint

mingjiecn:


src/routers/datatypeRouters/edges/variants_phenotypes.ts
```diff
@@ -12,7 +12,6 @@ import { TRPCError } from '@trpc/server'
 const MAX_PAGE_SIZE = 100
 
 const variantPhenotypeFormat = z.object({
-  'sequence variant': z.string().or(z.array(variantFormat)).optional(),
   phenotype_term: z.string().nullable(),
   study: z.string().or(z.array(studyFormat)).optional(),
```
shengchd, line 16: `study` could be an obj instead of an array as we did for other endpoints?

pedrohr (approved):

src/routers/datatypeRouters/edges/variants_phenotypes.ts
mingjiecn: good point. I was plan to change it but then forgot...

shengchd (approved):

PR #259, opened 2024-06-09
DSERV-497 New protein text search strategy

pedrohr:


src/routers/datatypeRouters/nodes/proteins.ts
```diff
@@ -82,6 +82,18 @@ async function findProteinsByTextSearch (input: paramsFormatType): Promise<any[]
...
+    FILTER ${getFilterStatements(proteinSchema, input)}
+    LIMIT ${input.page as number * limit}, ${limit}
+    RETURN { ${getDBReturnStatements(proteinSchema)} }
+  `
+
+  const exactObjects = await (await db.query(exactQuery)).all()
+  if (exactObjects.length !== 0) {
+    return exactObjects
+  }
+
```
mingjiecn, line 96: I feel this part is redundant compare to function findProteins. Can you refactor to remove the redundancy?

mingjiecn (changes requested):

src/routers/datatypeRouters/nodes/proteins.ts
pedrohr: just pushed a refactoring

mingjiecn (approved):

PR #262, opened 2024-06-13
DSERV-496 update variant filters for variant edge endpoint

mingjiecn:


src/routers/datatypeRouters/edges/variants_phenotypes.ts
```diff
@@ -150,90 +187,47 @@ async function getHyperedgeFromVariantQuery (input: paramsFormatType): Promise<a
...
 
 async function findPhenotypesFromVariantSearch (input: paramsFormatType): Promise<any[]> {
+  variantQueryValidation(input)
   delete input.organism
-  if (input.variant_id !== undefined) {
-    input._key = input.variant_id
-    delete input.variant_id
-  }
-
   if (input.funseq_description !== undefined) {
```
shengchd, line 231: This could be removed, we no longer have them in input?

src/routers/datatypeRouters/edges/variants_phenotypes.ts
```diff
@@ -150,90 +187,47 @@ async function getHyperedgeFromVariantQuery (input: paramsFormatType): Promise<a
...
+    FOR edgeRecord IN ${variantPhenotypeToStudy.db_collection_name as string}
+    FILTER edgeRecord._from IN primaryEdges ${hyperEdgeFilter.replaceAll('record', 'edgeRecord')}
+    SORT '_key'
+    LIMIT ${input.page as number * limit}, ${limit}
+    RETURN {
+      'study': ${input.verbose === 'true' ? `(${verboseQuery})[0]` : 'edgeRecord._to'},
+      ${getDBReturnStatements(variantPhenotypeToStudy).replaceAll('record', 'edgeRecord')}
+    }
+    `
+  } else {
```
shengchd, line 213: I'm thinking the condition `variantIDs.length > 0` there, so when you do a query for a non-existing variant in input, it could still return some results only based on the `log10pvalue` in input, which is a bit misleading?

src/routers/datatypeRouters/edges/variants_phenotypes.ts
mingjiecn: Will remove. Thanks!

src/routers/datatypeRouters/edges/variants_phenotypes.ts
mingjiecn: You are correct. I need to rethink about it.

shengchd (approved):

src/routers/datatypeRouters/edges/variants_phenotypes.ts
```diff
@@ -1,16 +1,31 @@
...
+  chr: z.string().trim().optional(),
+  position: z.string().trim().optional(),
+  phenotype_id: z.string().trim().optional(),
+  log10pvalue: z.string().trim().optional(),
+  organism: z.enum(['Homo sapiens']),
+  verbose: z.enum(['true', 'false']).default('false'),
+  page: z.number().default(0),
+  limit: z.number().optional()
+})
+
```
pedrohr, line 28: why didn't the extension of variantsQueryFormat from the original variants.ts file work?

src/routers/datatypeRouters/nodes/variants.ts
```diff
@@ -201,17 +201,49 @@ async function conditionalSearch (input: paramsFormatType): Promise<any[]> {
...
+  const query = `
+    FOR record IN ${variantSchema.db_collection_name as string} ${useIndex}
+    ${filterBy}
+    SORT record._key
+    LIMIT 0, ${QUERY_LIMIT}
+    RETURN record._id
+  `
+  return await (await db.query(query)).all()
+}
+
```
pedrohr, line 235: this is the equivalent of the `variantSearch` method, like:
`variant = variantSearch(input)[0].id`

src/routers/datatypeRouters/edges/variants_phenotypes.ts
mingjiecn: I like the idea to reuse the variantsQueryFormat from variants.ts. There are a couple issues for me: first you need to remove some fields and merge more objects to update the query format, which can be error-prone. Second, the order of the fields after remove and merge is not the order I want. The order of the params for each endpoint right now is inconsistent, we may need to figure a rule to decide the order of certain type of params.

src/routers/datatypeRouters/nodes/variants.ts
mingjiecn: try to test it but got an error: `Element implicitly has an 'any' type because expression of type '0' can't be used to index type 'Promise<any[]>'. Property '0' does not exist on type 'Promise<any[]>'.ts(7053)`

PR #271, opened 2024-07-03
DSERV-560-region-search

mingjiecn:


pedrohr (approved):

src/routers/datatypeRouters/_helpers.ts
```diff
@@ -13,7 +13,11 @@ export function validRegion (region: string): string[] | null {
     if (breakdown === null || breakdown.length < 4) {
       return null
     }
-    return breakdown
+    const start = parseInt(breakdown[2], 10)
+    const end = parseInt(breakdown[3], 10)
```
pedrohr, line 17: the base 10 is default for `parseInt`, you can omit it

pedrohr (changes requested):
all good! just a small lint, the base 10 is default for parseInt

pedrohr (approved):

PR #276, opened 2024-07-09
DSERV-515 genes_pathways name

pedrohr:


data/schema-config.yaml
```diff
@@ -1535,6 +1535,8 @@ gene to pathway association:
     from: gene
     to: pathway
   properties:
+    name: str
+    inverse_name: str
```
pedrohr, line 1539: this adapter uses biocypher, so we need to add to the schema.

Other adapters won't need it because we don't use `name` and `inverse_name` in the API.

mingjiecn (approved):

PR #285, opened 2024-07-25
DSERV-517 Adding edge labels for genes_transcripts and transcripts_proteins

pedrohr:


mingjiecn (changes requested):
since those adapters use biocypher, yaml schema should be also updated?

mingjiecn (approved):

PR #305, opened 2024-08-17
DSERV-595 genes/predictions endpoint

pedrohr:


src/routers/datatypeRouters/edges/genes_regulatory_regions.ts
```diff
@@ -0,0 +1,107 @@
...
+    },
+    predictions: regulatoryRegionGenes
+  }
+  console.log(stuff)
+  return stuff
+}
+
+const regulatoryRegionsFromGenes = publicProcedure
+  .meta({ openapi: { method: 'GET', path: '/genes/predictions', description: descriptions.genes_predictions } })
+  .input(singleGeneQueryFormat.merge(commonNodesParamsFormat).omit({ organism: true }))
```
mingjiecn, line 101: gene related edge query should have same filters based on the ticket here: https://igvf.atlassian.net/browse/DSERV-564. And the format is here: https://github.com/IGVF-DACC/igvf-catalog/blob/72a145a3454d1ce8da7bd083a8b99b7d2db4a197/src/routers/datatypeRouters/params.ts#L55. It also includes alias, unless the user want the specific gene filters?

src/routers/datatypeRouters/edges/genes_regulatory_regions.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (genes.length === 0) {
+    throw new TRPCError({
+      code: 'NOT_FOUND',
+      message: 'Gene not found.'
+    })
+  }
+
+  // assuming unique return
+  // input params are IDs only
+  const gene = genes[0]
```
mingjiecn, line 60: Sometime it still possible to get more than one gene based on the gene filters in genesCommonQueryFormat. I think alias may give you more than one gene sometimes.

mingjiecn (changes requested):

src/routers/datatypeRouters/edges/genes_regulatory_regions.ts
pedrohr: I don't think all gene endpoints should share the same query parameters. Not all endpoints expect to be related to multiple genes. This endpoint is a specific use-case for Daofeng's gene page.

`singleGeneQueryFormat` doesn't include alias:

`export const singleGeneQueryFormat = z.object({
  gene_id: z.string().trim().optional(),
  hgnc: z.string().trim().optional(),
  gene_name: z.string().trim().optional()
})`

They all should return an individual gene.

mingjiecn (approved):

mingjiecn (approved):

PR #309, opened 2024-08-26
DSERV-597-genes-structure

mingjiecn:


src/routers/datatypeRouters/nodes/genes_structure.ts
```diff
@@ -0,0 +1,139 @@
...
+  if (!fromProtein) {
+    const query = `
+      FOR record in ${genesStructureSchema.db_collection_name as string}
+      ${filterBy}
+      SORT record.gene_id, record.transcript_id, record['start:long']
+      LIMIT ${page as number * limit}, ${(page as number + 1) * limit}
+      RETURN {${getDBReturnStatements(genesStructureSchema)}}
+    `
+    return await (await db.query(query)).all()
+  } else {
```
shengchd, line 102: Am I understanding correctly that the protein query fields will override other fields then? Maybe it's fine, we can just make it clear in the endpoint description part.

src/routers/datatypeRouters/nodes/genes_structure.ts
```diff
@@ -0,0 +1,139 @@
...
+    const proteinInput = {
+      protein_id: input.protein_id,
+      protein_name: input.protein_name,
+      organism: input.organism,
+      page: 0,
+      limit: 25
```
shengchd, line 108: I'm wondering how this number of limit is set?

shengchd (commented):
Looks good overall, just have two questions on the protein query part.

src/routers/datatypeRouters/nodes/genes_structure.ts
mingjiecn: I updated the description.

src/routers/datatypeRouters/nodes/genes_structure.ts
mingjiecn: actually the limit is already set in the query function, so I will remove it. Thanks!

shengchd (approved):
looks good

PR #315, opened 2024-09-05
DSERV-594-write-to-s3

ottojolanki:


data/data_loader.py
```diff
@@ -1,22 +1,77 @@
...
+    key=non_adapter_signature_namespace.output_bucket_key,
+    session=boto3.Session(
+        profile_name=non_adapter_signature_namespace.aws_profile)
+)
 
-for a in adapters:
-    adapter = ADAPTERS[a]
-    adapter.write_file()
+adapter = LABEL_TO_ADAPTER[non_adapter_signature_namespace.adapter](
+    **vars(adapter_signature_namespace), writer=writer)
```
mingjiecn, line 76: really clever

mingjiecn:
Is the data_loader for sample loading only or for both sample loading and real data loading? Looks like we need to provide all the parameter values for sample loading as well? Is there a way that we don't need to provide those values, just need to provide the key from Adapters if it is sample loading since we don't usually change sample file when loading sample?

ottojolanki:
> Is the data_loader for sample loading only or for both sample loading and real data loading? Looks like we need to provide all the parameter values for sample loading as well? Is there a way that we don't need to provide those values, just need to provide the key from Adapters if it is sample loading since we don't usually change sample file when loading sample?

I think (if this does not bloat the repo too much) we might want to:

1. Write a script (bash probably) to parse all the sample data to jsonl and also store all the sample jsonls in repo.
2. Write a script (bash probably) to load sample jsonls to dev arango (using Loader class yet to be implemented).

So ideally the same (python) loader/parser scripts will be used for both dev and prod setups.

pedrohr:
The code looks good! Great job!

I double down on Mingjie's suggestion. The `ADAPTERS` dict with the instantiation of all adapters with sample data is not used. Should we call them automatically if no custom parameters are passed to the data_loader? We use them quite a lot during development and testing. Once the local format is ok, we can just run it by pointing to S3 to save the parsed data.

pedrohr:
We might also have to rename the data_loader.py script to something like data_parser.py or similar since we are not loading the data into any of the databases. The "new" data_loader would fetch the data from S3 or locally and insert them into ArangoDB or Clickhouse. What do you think?

ottojolanki:
Yea, the unused import is an oversight on my part. I also agree with the renaming. Will get those fixed and will also add the ontologies special case.

pedrohr:
Oh I mean we should keep the import, and use it if no parameters are passed, as default behavior of the loader.

ottojolanki:
What exactly is the "default behavior" here? Currently the `data_parser.py` (formerly known as `data_loader.py`) does not load anything. I think maybe if we want to keep the dev setup default behavior ACTIVE_ADAPTERS -style, we should do it in a separate dev setup parser script instead, to keep the main data_parser.py clean?

pedrohr:
Oh I see, yes that makes sense. I will move the default behavior of loading the sample data into the new script.

pedrohr (approved):

PR #320, opened 2024-09-13
DSERV-571 S3 Data ingestion script into ArangoDB

pedrohr:


data/adapters/storage.py
```diff
@@ -0,0 +1,126 @@
+import yaml
```
pedrohr, line 1: this is the old __init__.py with only the methods we currently need, we will continue to modify and refactor as more Clickhouse features get added.

mingjiecn (approved):

mingjiecn:
code looks good. I think we need to run the whole process with several adapters to test it.

PR #328, opened 2024-09-25
DSERV-619-complex

mingjiecn:


src/routers/datatypeRouters/edges/motifs_proteins.ts
```diff
@@ -47,30 +52,50 @@ async function proteinsFromMotifSearch (input: paramsFormatType): Promise<any[]>
...
     LET sources = (
       FOR record in ${motifSchema.db_collection_name as string}
       ${filterBy}
       RETURN record._id
     )
+    LET motifsProteins = (
 
     FOR record IN ${motifProteinSchema.db_collection_name as string}
-      FILTER record._from IN sources
+      FILTER record._from IN sources and record._to LIKE 'proteins/%'
```
pedrohr, line 75: does this perform well? I'm always afraid of LIKE in queries

mingjiecn: It seems fine on my local laptop. Probably because the dataset is not big? Only 480 edges... But I am not sure how big this collection will grow. If not efficient for big dataset, what should we do? Should we add some field to categorize different _to nodes so we can filter them easily?

pedrohr: yes, and we might have to add a separate index for that case if it doesn't perform well, but for now, it seems like it's ok

pedrohr (approved):

PR #329, opened 2024-09-27
DSERV-570 Clickhouse Data Ingestion

pedrohr:


data/db/clickhouse.py
```diff
@@ -0,0 +1,183 @@
...
+            if len(bulk) == 10000:
+                try:
+                    client.execute('INSERT INTO ' + collection +
+                                   ' (' + sql_properties + ') VALUES', bulk)
+                    bulk.clear()
+                except:
+                    # every collection has particular edge cases
+                    # this is needed until we have all collections loaded
+                    import pdb
+                    pdb.set_trace()
```
pedrohr, line 129: for now, I thought about leaving this until we ingest all collections into Clickhouse. Every dataset has its own small edge cases and we will have to adapt to each one of them.

I can also remove this for the PR, please let me know your thoughts.

data/schema-config.yaml
```diff
@@ -1082,26 +1082,23 @@ AFGR variant to gene expression association:
...
-      p_value: int
-      effect_size: int
-      label: str
-      source: str
-      source_url: str
-      name: str
-      inverse_name: str
-      biological_process: str
+  accessible_via:
+    name: e-qtls
```
mingjiecn, line 1086: unrelated question, what is name in accessible_via for?

data/db/clickhouse.py
```diff
@@ -0,0 +1,183 @@
...
+            else:
+                clickhouse_schema[table_name] = {'properties': properties}
+
+            if relationship:
+                clickhouse_schema[table_name]['relationship'] = {
+                    'from': relationship['from'],
+                    'to': relationship['to']
+                }
+
+        for s in clickhouse_schema:
```
mingjiecn, line 53: Can you explain what does the code to here for relationship?

data/adapters/ontologies_adapter.py
```diff
@@ -106,21 +103,9 @@ def process_file(self):
...
-
-        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
-            with urllib.request.urlopen(Ontology.ONTOLOGIES[self.ontology]) as response:
-                with gzip.GzipFile(fileobj=response) as uncompressed:
-                    file_content = uncompressed.read()
-                    temp_file.write(file_content)
-
-            temp_file_path = temp_file.name
-            onto = get_ontology(temp_file_path).load()
-
```
pedrohr, line 119: we are no longer going to download each ontology file on the fly. Each owl file will have to be passed individually

data/schema-config.yaml
```diff
@@ -687,8 +687,8 @@ variant to protein association:
         - label, source
         - source
   relationship:
-    from: sequence variant
-    to: protein
+    from: variants
+    to: proteins
```
mingjiecn, line 691: to can also be complex

data/schema-config.yaml
```diff
@@ -281,8 +281,8 @@ transcribed to:
   label_as_edge: TRANSCRIBED_TO
   db_collection_name: genes_transcripts
   relationship:
-    from: gene
-    to: transcript
+    from: genes
```
mingjiecn, line 284: can also from mm_genes to mm_transcripts

data/db/clickhouse.py
mingjiecn: sounds good

data/db/clickhouse.py
```diff
@@ -0,0 +1,184 @@
...
+            # clickhouse discourages the use of NULL, using '' instead
+            if value == None or value == []:
+                value = ''
+            processed_data.append(value)
+
+        processed_data.append(data['_key'])
+        if data.get('_from') is not None and data.get('_to') is not None:
+            # removing collection prefix from _from and _to. E.g. genes/ENSG00000148584 => ENSG00000148584
+            processed_data.append(data['_from'].split('/')[-1])
+            processed_data.append(data['_to'].split('/')[-1])
```
mingjiecn, line 99: how do you figure out which collection to go if there are more than one collection in from or to?

mingjiecn (changes requested):
Please see my comments. Thank you!

data/db/clickhouse.py
pedrohr: at this time, we don't need the combinatorial aspect, we just need to convert that list of tables into ID columns for Clickhouse. For example:

from: genes, mm_genes
to: transcripts, mm_transcripts

must be converted into:

genes_id String
mm_genes_id String
transcripts_id String
mm_transcripts_id String

as foreign keys in the SQL schema.

pedrohr: good catch, will look into this!

ottojolanki:
Some thoughts:
* It would probably make sense to put the most specific types (clickhouse) into the schema abstraction config yaml and make the transformation to less specific types when generating schema for ArangoDB.
* How difficult would it be to make the schema treatment similar between arango and clickhouse. What I mean by this is that both of those would be generated from the schema yaml using scripts and then stored in the repo as json/sql.
* Schema parsing/generation should be separated from data parsing/loading/ingestion for both databases.
* We should strive to use the bulk s3 ingestion for clickhouse. (see this blog series https://clickhouse.com/blog/getting-data-into-clickhouse-part-3-s3 for details about what I'm talking about)
* We need to add a way to define primary key columns.
* Unfortunately in some cases we may need to "manually" adjust clickhouse schemas (for example if we need to make some transformations like removing a `collection_name/` prefix from a value, get a nested value such as a frequency from the annotations for variants OR we need to adjust the parsing to contain these clickhouse specific values.

PR #367, opened 2025-02-05
DSERV-719-s3-tags

ottojolanki:


data/adapters/writer.py
```diff
@@ -46,6 +53,8 @@ def write(self, content):
 
     def close(self):
         self.s3_file.close()
+        print(self.version_tag)
```
mingjiecn, line 56: remove the print statement?

mingjiecn (approved):

data/adapters/writer.py
ottojolanki: Damn.

PR #368, opened 2025-02-05
DSERV-727 new acceptance/response time test script

pedrohr:


mingjiecn (approved):
Have a couple question:
1. if you run the script the second time. Will the response time be much shorter because of cache?
2. Should you also save the output files in repo too?

pedrohr:
1. Yes, DB caching will shorten response times.
2. We should, I just wanted to do it after v0.4.1b because we are fixing a few bugs from a few endpoints. After that, we should be good to save.

PR #371, opened 2025-02-06
DSERV-635-dey-predictions

ian-whaling:


shengchd (commented):
the adapter looks good, will need changes on other files:

- put an example data file under data/samples
- add the adapter to LABEL_TO_ADAPTER in active_adapters.py
- update schema-config.yaml (mainly the properties part)
- update data_sources.yaml (can wait until the data is in portal)
 @pedrohr can you also take a look to see if I'm missing anything? I suppose we are still keeping schema-config.yaml. There's already a 'variant to phenotype' entry there for GWAS data with different properties, should we add a new entry for 'variant to phenotype V2F' then?

PR #387, opened 2025-02-25
DSERV-759-file-set

ian-whaling:


data/adapters/helpers.py
```diff
@@ -128,3 +128,151 @@ def to_float(str):
...
+                    portal_url + input_file_set + '/@@object?format=json').json()
+                preferred_assay_titles.add(
+                    input_file_set_object.get('preferred_assay_title'))
+                assay_term = input_file_set_object.get('assay_term')
+                assay_term_object = requests.get(
+                    portal_url + assay_term + '/@@object?format=json').json()
+                assay_term_ids.add(assay_term_object.get('term_id'))
+    else:
+        raise(ValueError(
+            f'Loading data from file sets other than prediction sets and analysis sets is currently unsupported.'))
```
shengchd, line 204: For SEMpl (Boyle lab) we will also load their model files (as motifs) under model sets, but we can add the support later.

data/tests/test_helpers.py
```diff
@@ -69,3 +69,48 @@ def test_to_float_adapts_exponent_correctly():
...
+        'prediction': False,
+        'prediction_method': None,
+        'software': ['sceptre'],
+        'sample': 'CL:0000540',
+        'sample_id': sorted(['IGVFSM7750SNNY', 'IGVFSM8317ZTFV', 'IGVFSM8382KOXO', 'IGVFSM9913PXTT']),
+        'simple_sample_summary': 'neuron differentiated cell specimen',
+        'donor_id': 'IGVFDO1756PPKO',
+        'treatments_term_ids': None,
+        'publications': 'doi:10.1038/s41467-024-52490-4',
+    }
```
shengchd, line 109: can we add a test case where `treatment_term_ids` is not None? If no data current available we can add this later.

data/adapters/helpers.py
```diff
@@ -128,3 +128,151 @@ def to_float(str):
...
+                treatment_object = requests.get(
+                    portal_url + treatment + '/@@object?format=json').json()
+                treatment_ids.add(treatment_object.get('term_id'))
+                treatment_term_names.add(
+                    treatment_object.get('treatment_term_name'))
+                if 'publications' in treatment_object:
+                    publication_ids.update(get_publication_ids(
+                        portal_url, treatment_object))
+            treatment_term_names = ', '.join(list(treatment_term_names))
+            simple_sample_summary = f'{simple_sample_summary} treated with {treatment_term_names}'
```
shengchd, line 256: trying to think if we want treatment_terms_name or summary instead, the latter contains duration info. The encode E2G annotations are split by treatment + duration, so would be useful to have that in sample summary, but might not be true for IGVF analysis sets?

shengchd (commented):
Looks good! I just have some questions on the treatment term field. Also would be good to add an entry of files_filesets in schema-config.yaml (including props there) for record.

data/adapters/helpers.py
ian-whaling: left a note to add support later.

data/tests/test_helpers.py
ian-whaling: Done

data/adapters/helpers.py
ian-whaling: We do have duration, I can add that. The summary for treatments in IGVF also include amount & amount_units though.

PR #390, opened 2025-03-05
DSERV-771-deduplicate

ottojolanki:


mingjiecn (approved):

mingjiecn (changes requested):
since it generates vrs_digest to check deduplication we should use it to update _key as well. Need to update _key for human and mouse.

data/adapters/favor_adapter.py
```diff
@@ -224,7 +229,7 @@ def process_file(self):
                 hgvs = build_hgvs_from_spdi(spdi)
 
                 to_json = {
-                    '_key': id,
+                    '_key': allele_vrs_digest,
```
mingjiecn, line 232: Do we use spdi as _key if spdi is short enought?

ottojolanki: No.

ottojolanki: Right? Why would we?

mingjiecn (approved):

PR #394, opened 2025-03-08
DSERV-766-deduplication

ottojolanki:


ottojolanki:
Similarly we want to use the ReplacingMergeTree engine for any table we want to run some deduplication/deletes on. Not sure which ones these are gonna be?

PR #398, opened 2025-03-18
DSERV-809 Modifying CCRE ids

pedrohr:


data/adapters/ccre_adapter.py
```diff
@@ -54,11 +55,9 @@ def process_file(self):
             for row in reader:
                 try:
                     description = CCRE.BIOCHEMICAL_DESCRIPTION.get(row[9])
-                    # change to _id + filename?
-                    _id = row[3]
                     _props = {
-                        '_key': _id,
-                        'name': _id,
+                        '_key': build_regulatory_region_id(row[0], row[1], row[2]) + '_' + self.filename,
```
shengchd, line 59: seems we passed class_name for other genomic elements, so would add it: build_regulatory_region_id(row[0], row[1], row[2], 'candidate_cis_regulatory_element')

pedrohr: just added.

shengchd (approved):

PR #410, opened 2025-04-09
DSERV-648-protein-adapter-ENSP-key

shengchd:
- rewrite adapter for loading proteins nodes and edge collections using ENSP as keys instead of uniprot ids
- reconcile  info from gencode gtf file (ENSP &ENST ids) and uniprot dat file (mapping from uniprot -> ENSPs, protein names, and dbxrefs)
- script for generate pkl mapping files between ENSP <-> uniprot ids (in both ways)


data/adapters/gencode_protein_adapter.py
```diff
@@ -0,0 +1,232 @@
+# TODO:
+# check how to add isoform id to uniprot id
```
pedrohr, line 2: we can remove this TODO now :)

pedrohr (approved):

PR #411, opened 2025-04-10
DSERV-760-file-set-encode

ian-whaling:


data/adapters/file_fileset_adapter.py
```diff
@@ -0,0 +1,228 @@
...
+    def process_file(self):
+        self.writer.open()
+        if self.label == 'encode_file_fileset':
+            _props = query_fileset_files_props_encode(self.accession)
+            self.writer.write(json.dumps(_props) + '\n')
+
+        self.writer.close()
+
+
+def query_fileset_files_props_encode(accession):
```
pedrohr, line 35: I think we should break down this method so it's easier to read/modify later, maybe `parse_prediction()`, `parse_sample_and_treatment()`, ... returning parts of the final json

pedrohr (changes requested):
Just a quick refactoring suggestion for the `query_fileset_files_props_encode` method

data/adapters/file_fileset_adapter.py
```diff
@@ -0,0 +1,260 @@
...
+                treatment_ids.add(treatment_id)
+                treatment_term_names.add(treatment['treatment_term_name'])
+        if treatment_term_names:
+            treatment_term_names = ', '.join(
+                sorted(list(treatment_term_names)))
+            simple_sample_summary = f'{simple_sample_summary} treated with {treatment_term_names}'
+        simple_sample_summaries.add(simple_sample_summary)
+
+
+def query_fileset_files_props_encode(accession):
```
pedrohr, line 212: the new methods can be moved into the `FileFileSet` class, for example:

`def query_fileset_files_props_encode(self, accession):`

and the call would be: `self.query_fileset_files_props_encode(self.accession)` on line 29

pedrohr (approved):

pedrohr (approved):

PR #412, opened 2025-04-10
DSERV-743 DSERV-627 dbSNFP 5.1 upgrade

pedrohr:


data/adapters/dbSNFP_adapter.py
```diff
@@ -119,30 +103,39 @@ def long_data(pos):
...
-                        'source': 'dbSNFP 4.5a',
+                        'source': 'dbSNFP 5.1a',
                         'source_url': 'http://database.liulab.science/dbNSFP',
                         'name': 'codes for',
                         'inverse_name': 'encoded by',
                         'chr': data(0),
-                        'pos': long_data(1),
-                        'ref': data(2),  # 1-based
-                        'alt': data(3),
+                        'pos': long_data(1),  # 1-based
```
mingjiecn, line 133: why do we use 1 based coordinate? I thought we will always convert to 0 base?

mingjiecn (approved):

mingjiecn (approved):

data/adapters/dbSNFP_adapter.py
```diff
@@ -119,11 +104,27 @@ def long_data(pos):
...
+                gene_name = data(12)
+                transcript_id = data(14)
+                hgvsp = data(19)
+                hgvs = data(20)
+                aapos = long_data(11)
+
+                if hgvs is None:
+                    # basic format `chr:pos:ref:alt` to reuse hgvs builder method
+                    spdi = CHR_MAP['GRCh38'].get(
+                        data(0)) + ':' + str(data(1)) + ':' + data(2) + ':' + alt
```
shengchd, line 120: spdi uses 0-based coordinate, will need to do data(1) -1

data/adapters/dbSNFP_adapter.py
```diff
@@ -156,45 +157,42 @@ def long_data(pos):
...
                     }
                 else:
                     to_json = {
                         '_key': key,
                         'name': key,
                         'ref': data(4),
-                        'alt': data(5),
-                        'aapos': long_data(11),  # 1-based
-                        'gene_name': data(12),
+                        'alt': alt,
```
shengchd, line 168: want to load aaalt data(5)

data/adapters/dbSNFP_adapter.py
```diff
@@ -119,11 +104,27 @@ def long_data(pos):
                     except:
                         return None
 
-                # gene_name + transcript_id + hgvsp + hgvs + splicing (in case pos == -1)
-                key = data(12) + '_' + data(14) + '_' + \
-                    (data(23) or '') + '_' + (data(22) or '')
-
-                if long_data(11) == -1:
+                alt = data(3)
+                if alt == 'X' and 'Ter' in hgvsp:
```
shengchd, line 108: X can be in both in aaref and aaalt, will need to convert to * if either of them are X (no need checking if Ter in hgvsp)

data/adapters/dbSNFP_adapter.py
```diff
@@ -156,45 +157,42 @@ def long_data(pos):
...
                         'inverse_name': 'has variant',
-                        'source': 'dbSNFP 4.5a',
+                        'source': 'dbSNFP 5.1a',
                         'source_url': 'http://database.liulab.science/dbNSFP'
                     }
                 else:
                     to_json = {
                         '_key': key,
                         'name': key,
                         'ref': data(4),
```
shengchd, line 167: aaref also needs to convert to * if it's X

data/adapters/dbSNFP_adapter.py
pedrohr: even if it's 'X'? currently we will replace by * otherwise the value of data(5)

shengchd: oh it's correct then, replace by * if X

data/adapters/dbSNFP_adapter.py
```diff
@@ -119,11 +104,31 @@ def long_data(pos):
...
+                gene_name = data(12)
+                transcript_id = data(14)
+                hgvsp = data(19)
+                hgvs = data(20)
+                aapos = long_data(11)
+
+                if hgvs is None:
+                    # basic format `chr:pos:ref:alt` to reuse hgvs builder method
+                    spdi = CHR_MAP['GRCh38'].get(
+                        data(0)) + ':' + str(data(1) - 1) + ':' + data(2) + ':' + alt
```
shengchd, line 124: alt needs to be data(3) here

shengchd (changes requested):
just have two places alt was mixed up with aaalt, otherwise looks good

shengchd (approved):

PR #420, opened 2025-05-02
DSERV-831 Ensembl Proteins API changes

pedrohr:


data/schema-config.yaml
```diff
@@ -598,31 +598,35 @@ protein:
     query:
       type: persistent
       fields:
-        - name, full_name, organism
-        - full_name, organism
+        - dbxrefs[*].id, names[*], full_names[*], organism
```
mingjiecn, line 601: Can you add the index to index-config.yaml?

mingjiecn (approved):
Once index-config is updated then it should be good.

PR #430, opened 2025-05-16
DSERV-854-update-files-filesets

ian-whaling:


data/data_parser.py
```diff
@@ -57,8 +57,10 @@
                     help='The path to the dat file from uniprotKB TrEMBL.')
 parser.add_argument('--filepath', type=str,
                     help='The path to the input file.')
-parser.add_argument('--accession', type=str,
-                    help='ENCODE or IGVF file accession to fetch and parse data from.')
+parser.add_argument('--accessions', nargs='+', type=str,
+                    help='One or more ENCODE or IGVF file accessions to fetch and parse data from.')
+parser.add_argument('--replace', action='store_true',
```
shengchd, line 62: I assume we should use replace = false in most cases (maybe add a default = false?), and will need to remember to fix conflicting jsonl files on s3 if replace = true.

data/adapters/file_fileset_adapter.py
```diff
@@ -387,21 +464,92 @@ def query_fileset_files_props_igvf(self, accession):
...
+                phenotypic_feature_names = self.none_if_empty([
+                    pf['feature']['term_name']
+                    for pf in phenotypic_features
+                ])
+                ethnicities = self.none_if_empty(
+                    donor_object.get('ethnicities', []))
+
+            elif source == 'ENCODE':
+                health_status = donor_object.get('health_status')
+                phenotypic_feature_names = [
```
shengchd, line 520: have cases with health_status = unknown (e.g. ENCDO116AAA), should probably convert to None.
Also I feel it's not so useful to store ENOCDE phenotypes without an ontology term (e.g. ENCDO000ACV -> endometrial adenocarcinoma), might want to do some premapping for those cases.

data/adapters/file_fileset_adapter.py
```diff
@@ -69,24 +94,26 @@ def get_software_igvf(self, file_object, portal_url):
...
                 software_object = requests.get(
                     portal_url + software_version_object['software'] + '/@@object?format=json').json()
-                software.add(software_object['name'])
+                software.add(software_object['title'])
         return software
 
-    def parse_annotation(self, dataset_object, portal_url, prediction, prediction_method, software, preferred_assay_titles, assay_term_ids):
+    def parse_annotation(self, dataset_object, portal_url, class_type, method, software, preferred_assay_titles, assay_term_ids):
+        method = dataset_object['annotation_type']
         if 'prediction' in dataset_object['annotation_type']:
```
shengchd, line 102: this approach works most cases, but need to revisit manually curated caQTLs (ENCFF103XRK and ENCFF776CZZ), they should be 'experimental'.

data/adapters/file_fileset_adapter.py
```diff
@@ -311,43 +356,65 @@ def query_fileset_files_props_encode(self, accession):
...
+        else:
+            class_type = 'experimental'
         assay_term_ids, preferred_assay_titles = self.get_assay_encode(
             dataset_object, preferred_assay_titles, assay_term_ids)
+        if class_type == 'experimental':
+            if len(preferred_assay_titles) != 1:
+                raise (ValueError(
+                    f'Loading data from experimental data from multiple assays is unsupported.'))
+            else:
+                method = preferred_assay_titles[0]
```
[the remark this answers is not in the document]
shengchd: Another note on `software`: cCRE (ENCFF420VPZ) currently only has bed file converting tools in software. Will need to patch something like `SCREEN` (https://screen.encodeproject.org).

data/adapters/file_fileset_adapter.py
```diff
@@ -311,43 +356,65 @@ def query_fileset_files_props_encode(self, accession):
...
         preferred_assay_titles = set()
         assay_term_ids = set()
-        prediction = False
-        prediction_method = None
+        method = None
+        class_type = None
         if dataset_type == 'Annotation':
-            prediction, prediction_method = self.parse_annotation(
-                dataset_object, portal_url, prediction, prediction_method, software, preferred_assay_titles, assay_term_ids)
+            class_type, method = self.parse_annotation(
```
shengchd, line 362: it's a bit confusing that only `class_type` and `method` in return while the function also modifies some other parameters like `preferred_assay_titles`, but it works fine.

shengchd (commented):
Code looks good overall! Left some comments, and notes on datasets to revisit.

data/adapters/file_fileset_adapter.py
```diff
@@ -387,21 +464,92 @@ def query_fileset_files_props_igvf(self, accession):
...
+            yield {
+                '_key': accession,
+                'name': accession,
+                'sex': sex,
+                'age': age,
+                'age_units': age_units,
+                'ethnicities': self.none_if_empty(ethnicities),
+                'phenotypic_features': phenotypic_feature_ids,
+                'phenotypic_feature_names': phenotypic_feature_names,
+                'source': source
```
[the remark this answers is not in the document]
shengchd: another side note on donors: might need to be careful with common cell lines/donors like K562, we will have duplicated entries of them in encode and igvf (ENCDO000AAD, IGVFDO9208RPQQ)? but maybe it's fine

data/adapters/file_fileset_adapter.py
```diff
@@ -311,43 +356,65 @@ def query_fileset_files_props_encode(self, accession):
...
+        class_type = None
         if dataset_type == 'Annotation':
-            prediction, prediction_method = self.parse_annotation(
-                dataset_object, portal_url, prediction, prediction_method, software, preferred_assay_titles, assay_term_ids)
+            class_type, method = self.parse_annotation(
+                dataset_object, portal_url, class_type, method, software, preferred_assay_titles, assay_term_ids)
+            software_titles = ', '.join([software for software in software])
+            method = f'{method} using {software_titles}'
+        else:
+            class_type = 'experimental'
```
shengchd, line 367: really minor point: change to experiment as noun, to be more consistent with other two classes?

data/adapters/file_fileset_adapter.py
ian-whaling: mapping will be a bit of work, coz the health_status on ENCODE is free text and is not very curated.

data/adapters/file_fileset_adapter.py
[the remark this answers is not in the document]
ian-whaling: I think this should be done on ENCODE, not the script.

data/adapters/file_fileset_adapter.py
ian-whaling: I can have it return the rest of the properties.

data/adapters/file_fileset_adapter.py
ian-whaling: Ok.

data/adapters/file_fileset_adapter.py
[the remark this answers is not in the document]
shengchd: we can leave it for now

data/adapters/file_fileset_adapter.py
```diff
@@ -361,47 +438,124 @@ def query_fileset_files_props_igvf(self, accession):
...
+                    f"ontology_terms/{pf['feature']['term_id'].replace(':', '_')}"
+                    for pf in phenotypic_features
+                ])
+                ethnicities = self.none_if_empty(
+                    donor_object.get('ethnicities', []))
+
+            elif source == 'ENCODE':
+                ethnicities = donor_object.get('ethnicity', [])
+                phenotypic_feature_ids = None
+                ENCODE_disease_id_mapping = {
```
shengchd, line 522: i would probably move ENCODE_disease_id_mapping out of function, under class constants or somewhere else, for easier maintianance

shengchd (commented):
will need to check if the encode disease id mapping is complete

PR #431, opened 2025-05-19
DSERV-862-parsers-in-docker

ottojolanki:


docker/data/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+WORKDIR /data
+
+USER catalog
+
+RUN pip install --upgrade pip setuptools wheel
+
+RUN pip install -r requirements.txt
+
+RUN seqrepo pull -i 2024-12-20
```
mingjiecn, line 46: I think mouse data is also needed from seqrepo. Also not all parser need the sequence refs. Can some how make pulling seqrepo ref data optional?

mingjiecn (changes requested):
Please see my comment. Thanks

docker/data/Dockerfile
ottojolanki: This is done at build time, so it is not taking time when running. However it makes the docker image really large (12G with only human data), so I will try to figure out a way to serve seqrepo from s3.

PR #432, opened 2025-05-20
DSERV-857 adding name/inverse_name to edge responses

pedrohr:


mingjiecn (approved):

pedrohr:
Latest Changes:

1) Only return name

2) Return name := inverse_name if endpoint is opposite to the original collection name in ArangoDB, for example /genes/variants

PR #434, opened 2025-05-20
DSERV-866 script to tag jsonls based on catalog version

pedrohr:


ottojolanki (changes requested):
Functionality seems great. I think it would be good to add argumentparser, to make it more easy to change values.

ottojolanki (approved):

data/scripts/tag_s3_data.py
```diff
@@ -0,0 +1,95 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Append a value to the version tag of S3 objects.')
+    parser.add_argument('--bucket_name', required=True,
+                        help='The name of the S3 bucket.', default=BUCKET_NAME)
+    parser.add_argument('--append_value', required=True,
+                        help='The value to append to the version tag.', default=APPEND_VALUE)
+    parser.add_argument(
+        '--key_name', help='The key name for the version tag.', default=KEY_NAME)
+    parser.add_argument('--dry_run', action='store_true',
```
ottojolanki, line 89: Defaulting a `store_true` arg to `True` will never allow it to be `False`.

PR #441, opened 2025-06-04
DSERV-853 Adding files-filesets endpoint

pedrohr:


data/index-config.yaml
```diff
@@ -60,22 +60,16 @@ drugs:
...
-      - sample_ids
-      - simple_sample_summaries
-      - donors
-      - treatments_term_ids
-      - publication
-      - source
+      - donors[*]
+      - samples[*]
+      - simple_sample_summaries[*]
+      - software[*]
```
shengchd, line 72: also need source and class

PR #449, opened 2025-06-25
DSERV-892 Ignore '*' alt and ref

pedrohr:


data/adapters/AFGR_eqtl_adapter.py
```diff
@@ -42,6 +42,10 @@ def process_file(self):
             for row in qtl_csv:
                 chr, pos, ref, alt = row[6].split('_')
 
+                # handling deletions:
+                if alt == '*':
+                    alt = 'T'
```
mingjiecn, line 47: why replace * with T? I thought for deletion, alt is empty string?

PR #453, opened 2025-07-01
DSERV-841-variant-effects-2

ian-whaling:


shengchd (commented):
See comment on edge loading part

data/adapters/VARIANT_EFFECTS_variant_gene_adapter.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if skipped_spdis:
+            print(f'Skipped {len(skipped_spdis)} variants:')
+            for skipped in skipped_spdis:
+                print(f"  - {skipped['variant_id']}: {skipped['reason']}")
+            with open('./skipped_variants.jsonl', 'a') as out:
+                for skipped in skipped_spdis:
+                    out.write(json.dumps(skipped) + '\n')
+
+        loaded_variants = bulk_check_spdis_in_arangodb(spdi_to_variant.keys())
```
shengchd, line 98: I would probably move this forward, before load_variant function, because you don't need to validate + normalize variant if it's already loaded. It's depending on the data also - if the majority is new variants then it's fine to leave it here

data/adapters/VARIANT_EFFECTS_variant_gene_adapter.py
```diff
@@ -0,0 +1,142 @@
...
+                continue
+            else:
+                variant.update({
+                    'source': self.SOURCE,
+                    'source_url': self.source_url,
+                    'files_filesets': f'files_filesets/{self.file_accession}'
+                })
+                self.writer.write(json.dumps(variant) + '\n')
+
+    def process_edge(self, spdi_to_row, loaded_variants):
```
shengchd, line 117: think you want to only skip variants with 'skipped messages'

data/adapters/VARIANT_EFFECTS_variant_gene_adapter.py
ian-whaling: But I need to get the normalized SPDI to check the database with?

PR #458, opened 2025-07-07
DSERV-901-mv-variant-loader-2

shengchd:


data/adapters/helpers.py
```diff
@@ -416,11 +418,112 @@ def get_ref_seq_by_spdi(spdi, species='human'):
...
+        _id = allele.digest
+
+    variation_type = 'SNP'  # should be SNV more broadly
+    if len(ref) < len(alt):
+        variation_type = 'insertion'
+    elif len(ref) > len(alt):
+        variation_type = 'deletion'
+    elif len(ref) > 1:
+        # e.g. NC_000018.10:31546003:AA:TG
+        variation_type = 'deletion-insertion'
```
ian-whaling, line 502: What about 
elif len(ref) == len(alt) == 1:
    variation_type = 'substitution'

shengchd: so replace SNP with substitution? I wanted to keep it the same as those we already loaded from favor for now

ian-whaling: Ok that's fine, I forgot there was a default value

data/tests/test_helpers.py
```diff
@@ -133,3 +133,136 @@ def test_bulk_check_spdis_in_arangodb():
...
+    variant_id = 'NC_000001.11:12345:A:N'
+    result, skipped = load_variant(variant_id)
+    assert result == {}
+    assert skipped == {'variant_id': variant_id,
+                       'reason': 'Ambigious alt allele'}
+
+
+@patch('adapters.helpers.get_ref_seq_by_spdi', return_value='T')
+def test_spdi_empty_ref(mock_ref_seq):
+    variant_id = 'NC_000010.11:79347444::CCTCCTCAGG'
```
shengchd, line 223: I think this should be accepted, return_value for get_ref_seq_by_spdi should be ''

ian-whaling: fixed

PR #467, opened 2025-07-23
DSERV-883 All-scores for gene/dataset endpoint 

pedrohr:


pedrohr:
I didn't add pagination by Shane's request. We will run a few more tests when more data is loaded and we will add pagination if needed.

PR #499, opened 2025-08-14
DSERV-843-gersbach-bulk-CRISPR-2

ian-whaling:


data/adapters/gersbach_E2G_CRISPR_adapter.py
```diff
@@ -108,6 +161,7 @@ def process_file(self):
                         'biological_context': biosample_term,
```
shengchd, line 161: is this an array right now? think we agreed to have this field as str if possible

ian-whaling: You're right, I can adjust it to a string. I think the Perturb-seq data may have to be reloaded then.

PR #532, opened 2025-09-30
DSERV-1004-terms-gene-products2

mingjiecn:


src/routers/datatypeRouters/edges/go_terms_annotations.ts
```diff
@@ -167,29 +163,28 @@ async function annotationsSearch (input: paramsFormatType): Promise<any[]> {
...
-      SORT record._to
+      SORT record._from
       LIMIT ${page * limit}, ${limit}
 
       RETURN DISTINCT {
-        'name': record.name,
-        'annotation_id': targetReturn._id OR dbxrefTargetReturn._id,
-        'annotation_name': targetReturn.names[0] OR dbxrefTargetReturn.names[0],
-        'go_term_name': sourceReturn.name,
+        'name': record.inverse_name,
```
pedrohr, line 170: should we fix the `name` in the adapter so we can continue with a consistent `record.name`?

pedrohr: oh I see, that makes sense

pedrohr (approved):

PR #533, opened 2025-10-01
DSERV-1032 adding index for `score`

pedrohr:


data/index-config.yaml
```diff
@@ -276,8 +276,8 @@ genomic_elements_biosamples:
 genomic_elements_genes:
   persistent:
     fields:
-      - file_accession, source
```
pedrohr, line 279: `file_accession` (old files_filessets) doesn't exist anymore

PR #535, opened 2025-10-01
DSERV-1007-gwas

mingjiecn:


shengchd (commented):
I think `variants_to_genes` in input parameter can be removed as well.
I will archive the data file on portal (IGVFFI5724MMHI), the relevant line in data_sources.yaml can also be dropped.

shengchd (approved):

PR #543, opened 2025-10-07
DSERV-1069-downgrade-cdk

ottojolanki:


mingjiecn:
Do we need downgrade here: https://github.com/IGVF-DACC/igvf-catalog/blob/72f58e07bb48f7578f37200bbf2d2cd37adb2cc0/cdk_swagger/infrastructure/config.py#L97 ?

mingjiecn (approved):

PR #600, opened 2026-01-07
DSERV-990-mane-select

mingjiecn:


data/schemas/nodes/mm_transcripts.Gencode.json
```diff
@@ -95,7 +101,7 @@
   "additionalProperties": false,
   "accessible_via": {
     "filter_by_range": "start, end",
-    "return": "_id, chr, start, end, strand, gene_name, transcript_id, name, transcript_type, source, version, source_url, organism"
+    "return": "_id, chr, start, end, strand, gene_name, transcript_id, name, transcript_type, MANE_Select, source, version, source_url, organism"
```
shengchd: it might make more sense to drop this property from mm_transcripts, and not include in return since MANE_Select is only applicable to human transcripts. But I'm not sure if it would cause any issue in endpoint when we have a mix case from human & mouse in return?

pedrohr (approved):

PR #606, opened 2026-01-26
DSERV-1152-ADASTRA

mingjiecn:


data/adapters/adastra_asb_adapter.py
```diff
@@ -107,69 +104,47 @@ def process_file(self):
...
-                                    'inverse_name': 'has measurement'
-                                }
-
-                                if self.validate:
-                                    self.validate_doc(props)
-
-                                self.writer.write(json.dumps(props))
-                                self.writer.write('\n')
+                        for ensembl_id in ensembl_ids:
+                            # create edges in variants_proteins regardless of cell type
```
shengchd: the comments here can be removed

data/adapters/adastra_asb_adapter.py
```diff
@@ -107,69 +104,47 @@ def process_file(self):
...
+                            # the redundance will be resolved when importing into arangodb
+                            _key = variant_id + '_' + \
+                                ensembl_id + '_' + \
+                                row[21].replace(' ', '_') + \
+                                '_ADASTRA_' + cell_gtrd_id
```
shengchd, line 114: I checked the cell_gtrd_id in ADASTRA_cell_ontologies_mapped_ids.tsv they should be unique, can you double check when you reloading the collection?
Also I believe 'row[21].replace(' ', '_')' can be dropped from ID generation, shouldn't affect the uniqueness but fine to leave it there

shengchd (approved):
looks good just minor suggestions
I also left a note on 'score' in DSERV-1126, but it's for endpoint

data/adapters/adastra_asb_adapter.py
mingjiecn: 'row[21].replace(' ', '_')' is necessary since we don't want space in key. There are duplication because we have those two files: **CREB1_HUMAN@HepG2__hepatoblastoma_.tsv** and **CREB1_HUMAN@HepG2__hepatoblastoma__test.tsv**. Both of them match to ADASTRA id 430. It seems to me that we should remove the test file from our source data.

shengchd: I meant the info from row[21]like 'No Hit' should be unique to every variant - protein combination, so could be excluded from key generation since we already have variant and protein id there.

PR #619, opened 2026-02-13
DSERV-1180-COXPRESdb

mingjiecn:


data/adapters/coxpresdb_adapter.py
```diff
@@ -13,9 +14,12 @@
 
 class Coxpresdb(BaseAdapter):
     ALLOWED_LABELS = ['coxpresdb']
+    FILE_ACCESSION = 'IGVFFI3321YNBP'
```
shengchd, line 17: I feel this should be read from input para filepath instead of hard coded here? in case we have another version of file with a different accession in future
maybe more important for adapter has multiple files to load, like eqtl_catalog, this adapter is fine

shengchd (approved):

PR #621, opened 2026-02-20
DSERV-1185-files-filesets

mingjiecn:


pedrohr (approved):

data/scripts/generate_files_filesets.py
```diff
@@ -0,0 +1,106 @@
...
+    output_dir = Path(args.output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    with data_sources_path.open('r') as handle:
+        data_sources = yaml.safe_load(handle)
+
+    timestamp = datetime.now(timezone.utc).strftime('%Y%m%dT%H%M%SZ')
+    labels = {
+        'igvf_file_fileset': f'igvf_files_filesets_{timestamp}.jsonl',
+        'encode_file_fileset': f'encode_files_filesets_{timestamp}.jsonl'
```
shengchd: @mingjiecn can you add igvf_donor.jsonl and encode_donor.jsonl as well? think they should be updated simultaneously

ian-whaling: The adapter also writes novel (unloaded) sample terms (even if they are rare), may want to also include writing a file for that.

PR #624, opened 2026-02-24
DSERV-1174-qtl

mingjiecn:


data/adapters/eqtl_catalog_adapter.py
shengchd, line 157: I think we can drop this field, not sure why we included before

shengchd (approved):
code looks good, have a comment on a field could be dropped

PR #630, opened 2026-03-03
DSERV-1172 adding class/method/label for GWAS

pedrohr:


data/adapters/gwas_adapter.py
```diff
@@ -175,7 +175,10 @@ def process_variants_phenotypes(self, row):
             'source': 'OpenTargets',
             'version': 'October 2022 (22.10)',
             'name': 'associated with',
-            'inverse_name': 'associated with'
+            'inverse_name': 'associated with',
+            'class': 'observed data',
+            'method': 'GWAS',
+            'label': 'GWAS'
```
mingjiecn: need to get class and method value from igvf portal. I have a merged pr did similar work: https://github.com/IGVF-DACC/igvf-catalog/pull/624

data/schemas/edges/variants_phenotypes.GWAS.json
```diff
@@ -61,6 +61,27 @@
       "enum": [
         "associated with"
       ]
+    },
```
mingjiecn: Now since we added class, method and label, we can use base edge schema inside. I have a merged pr did similar work: https://github.com/IGVF-DACC/igvf-catalog/pull/624.

mingjiecn (changes requested):

data/schemas/edges/edges.base.json
```diff
@@ -53,5 +53,5 @@
       "source",
       "source_url"
     ],
-    "additionalProperties": false
+    "additionalProperties": true
```
mingjiecn: additionalProperties should be false?

data/adapters/gwas_adapter.py
```diff
@@ -173,9 +178,13 @@ def process_variants_phenotypes(self, row):
             '_key': key,
             'equivalent_ontology_term': equivalent_term_id,
             'source': 'OpenTargets',
+            'source_url': self.SOURCE_URL,
             'version': 'October 2022 (22.10)',
             'name': 'associated with',
-            'inverse_name': 'associated with'
+            'inverse_name': 'associated with',
+            'class': 'observed data',
```
mingjiecn: class value should be self.collection_class. Label value is the same with self.method

data/schemas/edges/variants_phenotypes.GWAS.json
```diff
@@ -3,75 +3,31 @@
   "$id": "edges/variants_phenotypes.GWAS.json",
   "db_collection_name": "variants_phenotypes",
   "type": "object",
+  "allOf": [
+    { "$ref": "edges.base.json" }
```
mingjiecn: should add other fields for example. properties and required inside allOf

mingjiecn (changes requested):

data/schemas/edges/edges.base.json
pedrohr: by being false, we can't modify the child classes... we can leave as false too and just import a subset of the properties

mingjiecn (approved):

PR #632, opened 2026-03-05
DSERV-1162 DSERV-1189 adding uniprot_ids to protein response across catalog APIs

pedrohr:


mingjiecn (changes requested):
I think all protein related APIs should be updated to return uniprot_ids.

pedrohr:
> I think all protein related APIs should be updated to return uniprot_ids.

all of them are, they all inherit the protein object from `proteins.ts`

mingjiecn:
> > I think all protein related APIs should be updated to return uniprot_ids.
> 
> all of them are, they all inherit the protein object from `proteins.ts`

Oh, my bad.

mingjiecn (approved):

PR #633, opened 2026-03-05
DSERV-1192-upgrade-cdk

ottojolanki:


mingjiecn (commented):
In the instruction we need to install this version:
npm install -g aws-cdk@2.151.0

Is this need to be changed?

mingjiecn (approved):

PR #638, opened 2026-03-14
DSERV-1196-remodel-MPRA

ian-whaling:


data/index-config.yaml
```diff
@@ -273,8 +273,8 @@ genomic_elements:
 genomic_elements_biosamples:
   persistent:
     fields:
-      - type, element_name, source, source_url, files_filesets
-      - element_name, source, source_url, files_filesets
+      - type, source, source_url, files_filesets
+      - source, source_url, files_filesets
       - source, source_url, files_filesets
```
pedrohr: duplicated line

pedrohr (changes requested):
just one duplicated line

data/index-config.yaml
ian-whaling: removed duplicate line

pedrohr (approved):

PR #642, opened 2026-03-24
DSERV-1204-remove-threshold-STARR-seq

ian-whaling:


shengchd (approved):

data/data_sources/data_sources.yaml
```diff
@@ -1081,6 +1081,7 @@ variant to biosample starr-seq:
     - https://api.data.igvf.org/tabular-files/IGVFFI5057AAGP/@@download/IGVFFI5057AAGP.bed.gz
   source: IGVF
   label: variant effect on regulatory element activity
+  threshold: 0.9
```
shengchd, line 1084: Not sure where we should document this constant - should we put it under edges schema or load as a field under files_filesets? @mingjiecn @ian-whaling  
It was a temporary approach to include the value here for generating the catalog file spreadsheet before

ian-whaling: idk I just saw it in other entries so I added it.

shengchd: yeah we can discuss it later, loading code is fine here

PR #650, opened 2026-03-31
DSERV-1216 renaming beta -> effect_size

pedrohr:


mingjiecn (changes requested):
Tests needs to be updated as well

mingjiecn (approved):

PR #676, opened 2026-04-22
DSERV-1248-crispr-modality

ian-whaling:


data/schemas/edges/genomic_elements_genes.ENCODE2GCRISPR.json
```diff
@@ -48,7 +48,18 @@
...
+          ]
+        },
+        "crispr_modality": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "description": "The CRISPR modality from the associated files_filesets record",
+          "enum": [
+            null,
```
shengchd: should all ENCODE CRISPR edge = 'interference'?

data/schemas/edges/variants_genes.VariantEFFECTSAdapter.json
```diff
@@ -93,6 +93,19 @@
             "Variant-EFFECTS"
           ]
         },
+        "crispr_modality": {
+          "type": [
+            "string",
+            "null"
```
shengchd: if it can be null, then should not be in required list (line 144)?

data/schemas/nodes/files_filesets.FileFileSet.json
```diff
@@ -265,6 +264,19 @@
             "pluripotent epiblast cell",
             "surface ectodermal cell"
           ]
+        },
+        "crispr_modality": {
+          "type": [
+            "string",
+            "null"
```
shengchd, line 275: same here

data/index-config.yaml
```diff
@@ -289,14 +290,18 @@ genomic_elements_genes:
   persistent:
     fields:
       - method
+      - crispr_modality
```
shengchd: do we need crispr_modality in genomic_elements/genes api query field? seems it's only added to the output field. I think index only needed if it's a query field. @mingjiecn @pedrohr

data/index-config.yaml
```diff
@@ -387,6 +392,7 @@ variants_genes:
       - biological_context, label, source, files_filesets
       - label, source, files_filesets
       - source, files_filesets
+      - crispr_modality, files_filesets
```
shengchd: same here

src/routers/datatypeRouters/nodes/files_filesets.ts
shengchd, line 23: can add crispr_modality to query field under files_filesets/, since it's a small collection

data/schemas/edges/genomic_elements_genes.ENCODE2GCRISPR.json
ian-whaling: yes

data/schemas/edges/variants_genes.VariantEFFECTSAdapter.json
ian-whaling: fixed, it should not be null. Nor should it be anything other than prime editing atm

data/schemas/edges/genomic_elements_genes.ENCODE2GCRISPR.json
shengchd: need to drop null then

data/schemas/nodes/files_filesets.FileFileSet.json
ian-whaling: "if it can be null, then should not be in required list?"
cell_annotation can be null and it is required. likewise with other properties?

data/index-config.yaml
mingjiecn: It can be useful for QA purpose.

data/schemas/nodes/files_filesets.FileFileSet.json
shengchd: oh I see in files_filesets optional fields are all specified this way, fine to leave then

data/index-config.yaml
shengchd: should we drop those after QA? especially for large collections like genomic_elements_genes

shengchd (approved):

src/routers/datatypeRouters/nodes/files_filesets.ts
ian-whaling: added

data/index-config.yaml
mingjiecn: yes if it is no other use. If we don't need it for QA, we can also drop it here

pedrohr: if it's not a query field then we don't need... ok to have it for QA then drop it

data/index-config.yaml
pedrohr: if it's not a query field then we don't need... ok to have it for QA then drop it

PR #683, opened 2026-04-24
DSERV-1231 removing 'filter' attribute from variants

pedrohr:


mingjiecn (changes requested):
Need to update variantFormat in API as well

mingjiecn (approved):

PR #688, opened 2026-04-29
DSERV-1262-genes-qtls

mingjiecn:


src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+  gene: z.object({
+    name: z.string(),
+    id: z.string()
+  }).nullable(),
+  genomic_element: z.object({
+    name: z.string(),
+    chr: z.string(),
+    start: z.number(),
+    stop: z.number(),
+    type: z.string()
```
pedrohr, line 37: do you need to return files_filesets?

src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+  return {
+    filterClause: clauses.length > 0 ? `AND ${clauses.join(' AND ')}` : '',
+    bindVars
+  }
+}
+
+function qtlReturnObject (geneExpr: string, genomicElementExpr: string): string {
+  return `{
+    _key: record._key,
+    variant: { chr: variant.chr, pos: variant.pos, SPDI: variant.spdi, rsid: variant.rsid },
```
pedrohr: SPDI should be lower case to match format

src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+    gene: ${geneExpr},
+    genomic_element: ${genomicElementExpr},
+    source: record.source,
+    method: record.method,
+    regulatory_type: record.regulatory_type,
+    gene_consequence: record.gene_consequence,
+    biological_context: record.biological_context,
+    nLog10Pvalue: record.log10pvalue,
```
pedrohr: I would double-check with Ian; we were trying to normalize this field... I think it's `log10pvalue` as other endpoints

src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+    study: record.study
+  }`
+}
+
+function variantsGenesQuery (filterClause: string, withLimit: boolean): string {
+  return `
+  FOR record IN variants_genes
+  FILTER record._to IN @geneIDs ${filterClause}
+  LET variant = DOCUMENT(record._from)
+  LET geneRecord = DOCUMENT(record._to)
```
pedrohr: move these two LET's below the LIMIT so it only materializes these variables after the cut, it optimizes the speed a bit

src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+  RETURN ${qtlReturnObject('{ name: geneRecord.name, id: geneRecord._key }', 'null')}
+  `
+}
+
+function variantsProteinsQuery (filterClause: string, withLimit: boolean): string {
+  return `
+  FOR record IN variants_proteins
+  FILTER record.gene IN @geneIDs ${filterClause}
+  LET variant = DOCUMENT(record._from)
+  LET geneRecord = DOCUMENT(record.gene)
```
pedrohr: moving LET below LIMIT here as well

src/routers/datatypeRouters/edges/genes_qtls.ts
```diff
@@ -0,0 +1,345 @@
...
+  LET genomicElement = DOCUMENT(record._to)
+  LET geneID = FIRST(
+    FOR pair IN genomicElementGenePairs
+    FILTER pair.element_id == record._to
+    RETURN pair.gene_id
+  )
+  LET geneRecord = DOCUMENT(geneID)
+  FILTER geneRecord != null
+  SORT record._key
+  ${withLimit ? 'LIMIT @offset, @limit' : ''}
```
pedrohr: move SORT + LIMIT clauses above LET's

src/routers/datatypeRouters/edges/genes_qtls.ts
mingjiecn: good idea

src/routers/datatypeRouters/edges/genes_qtls.ts
mingjiecn: Talked with Ben and Ian. Will use `nlog10pvalue` here. Also created a ticket to check this fields for all APIs

pedrohr (approved):

pedrohr (approved):

PR #691, opened 2026-05-05
DSERV-1208-gencc

mingjiecn:


data/adapters/helpers.py
```diff
@@ -788,3 +788,21 @@ def get_file_fileset_by_accession_in_arangodb(accession):
     db = ArangoDB().get_igvf_connection()
     files_filesets_collection = db.collection('files_filesets')
     return files_filesets_collection.get(accession)
+
+
+def get_gene_map_from_arangodb(field):
+    db = ArangoDB().get_igvf_connection()
+    cursor = db.aql.execute(
+        f'FOR gene IN genes RETURN {{ key: gene._key, value: gene.{field} }}'
```
pedrohr, line 796: does this work well without pagination?

pedrohr (approved):

data/adapters/helpers.py
mingjiecn: It works well for small collections. Python driver still fetch data in batches.

data/adapters/gencc_diseases_genes_adapter.py
```diff
@@ -0,0 +1,79 @@
...
+    def _get_collection_name(self):
+        return 'diseases_genes'
+
+    def process_file(self):
+        self.writer.open()
+        self.gene_map = get_gene_map_from_arangodb('hgnc')
+        if 'HGNC:32925' not in self.gene_map:
+            self.gene_map['HGNC:32925'] = ['ENSG00000288330']
+        # read the tsv file
+        with open(self.filepath, 'r', encoding='utf-8', newline='') as f:
```
shengchd, line 29: will need to post this tsv data file on portal and add to data_source.yaml

data/schemas/edges/diseases_genes.GenccDiseasesGenes.json
```diff
@@ -0,0 +1,174 @@
...
+        "Laboratory for Molecular Medicine",
+        "LiferaOmics",
+        "Myriad Women's Health",
+        "Orphanet",
+        "PanelApp Australia",
+        "Stanford Center for Undiagnosed Diseases",
+        "University of Washington Center for Rare Disease Research (UW-CRDR)"
+      ]
+    },
+    "pmid": {
```
shengchd: should be in plural form pmids if it's an array. also it's defined in format like 'http://pubmed.ncbi.nlm.nih.gov/33660365' in genes_genes.biogrid, should probably be in the same format here.

mingjiecn: I use pmid because this is how it defined in diseases_genes from Orphanet data

PR #692, opened 2026-05-06
DSERV-1271-nlog10pval

mingjiecn:


pedrohr (approved):
all looks good! 
I feel like it's a good idea to change the adapters in the future so we don't need extra conditionals on the API side

PR #693, opened 2026-05-06
DSERV-1264 adding shared security group for dev demos

pedrohr:


hitz (approved):
seems fine

ottojolanki:
It would be better to put the security group id into the config.py instead of hardcoding.

cdk_swagger/infrastructure/stacks/frontend.py
```diff
@@ -25,6 +26,12 @@ def __init__(
             **kwargs: Any
     ) -> None:
         super().__init__(scope, construct_id, **kwargs)
+        shared_sg = SecurityGroup.from_security_group_id(
```
ottojolanki, line 29: We cannot hardcode here, because this might get deployed do different accounts (prod, dev etc), I think we need to create elsewhere and pass in.

ottojolanki (approved):

PR #694, opened 2026-05-07
DSERV-1276-favor-use-helper

ottojolanki:


data/adapters/favor_adapter.py
```diff
@@ -184,88 +183,72 @@ def parse_metadata(self, info):
...
+
+            data_line = line.strip().split()
+
+            # data files sometimes add 'chr' before the chromosome value and sometimes they do not, normalizing it:
+            chrm = data_line[0].replace('chr', '')
+            ref = data_line[3]
+            alt = data_line[4]
+
+            variant_id = f'{chrm}-{data_line[1]}-{ref}-{alt}'
+            variant_json, skipped_message = load_variant(
```
mingjiecn, line 206: You will need to call get_seqrepo every time when calling load_variant. I feel this may slow down the parsing speed since we need to parse a tons of variants.

mingjiecn (approved):

data/adapters/favor_adapter.py
pedrohr: if you pass `translator` and `seq_repo` to the method, it won't call it again, so it shouldn't affect speed

PR #710, opened 2026-05-20
DSERV-1298-eve-mcap-score-delete

ottojolanki:


pedrohr (changes requested):
I think they should be added to properties; it’s part of the API return https://github.com/IGVF-DACC/igvf-catalog/blob/dev/src/routers/datatypeRouters/nodes/coding_variants.ts#L52-L62

ottojolanki:
> I think they should be added to properties; it’s part of the API return https://github.com/IGVF-DACC/igvf-catalog/blob/dev/src/routers/datatypeRouters/nodes/coding_variants.ts#L52-L62

None of the adapters create these fields. So we actually want to remove also from the API return?

pedrohr (approved):

PR #711, opened 2026-05-21
DSERV-1301-MPRAAdapter

mingjiecn:

PR #712, opened 2026-05-22
DSERV-1302-variant-biosample

mingjiecn:


pedrohr (approved):

ian-whaling (changes requested):
If you're changing minusLog10PValue name ought we also change minusLog10QValue to match the changes?

mingjiecn:
> If you're changing minusLog10PValue name ought we also change minusLog10QValue to match the changes?

Do you know how to update this field? What are the other fields that mean the same thing as minusLog10QValue? I can update them together.

ian-whaling:
> > If you're changing minusLog10PValue name ought we also change minusLog10QValue to match the changes?
> 
> Do you know how to update this field? What are the other fields that mean the same thing as minusLog10QValue? I can update them together.

It is an fdr

mingjiecn:
> > > If you're changing minusLog10PValue name ought we also change minusLog10QValue to match the changes?
> > 
> > 
> > Do you know how to update this field? What are the other fields that mean the same thing as minusLog10QValue? I can update them together.
> 
> It is an fdr

To keep things consistent I can rename it to neg_log_qvalue.  Rename it to fdr is kind of concerning. Since we don't know what kind fdr it is anymore. adjusted p value is also a fdr. I feel at least we need to distinguish neg_log_qvalue and adjusted p value in database. We can use fdr in APIs if needed.

data/adapters/mpra_adapter.py
ian-whaling, line 546: They are on the genomic_element_biosamples as well

ian-whaling: nvm handled in https://github.com/IGVF-DACC/igvf-catalog/pull/711

ian-whaling (approved):

data/schemas/edges/variants_biosamples.MPRAAdapter.json
ian-whaling: "description": "True if this edge met the significance threshold (e.g. neg_log10_qvalue >= 1)."

ian-whaling (changes requested):

ian-whaling (approved):

PR #713, opened 2026-05-22
DSERV-1291 Scorpion adapter

pedrohr:


data/adapters/scorpion_adapter.py
```diff
@@ -0,0 +1,66 @@
...
+        self.collection_class = file_metadata['catalog_class']
+        self.method = file_metadata['catalog_method']
+
+        with gzip.open(self.filepath, 'rt') as data_file:
+            data_csv = csv.DictReader(
+                data_file, delimiter='\t', fieldnames=self.filepath)
+
+            for row in data_csv:
+                props = {
+                    '_key': row['tf_ensembl'] + '_' + row['target_ensembl'] + '_' + self.LABEL,
```
mingjiecn: self.LABEL has space in it, not sure if this is good for key generation. Need to replace space, or use other string for example file accession?

data/data_sources/data_sources_file_fileset.yaml
```diff
@@ -29,6 +29,8 @@ igvf donor:
     --replace
 
 igvf_file_accessions:
+  Scorpion:
+    - IGVFFI3035IXUW
```
mingjiecn, line 33: Need to also merge in dev branch and generate new files_filesets data file for IGVF source.

mingjiecn (changes requested):

data/adapters/scorpion_adapter.py
pedrohr: ohh great catch... will replace by underline

data/data_sources/data_sources_file_fileset.yaml
mingjiecn: may need to regenerate it again since we don't know which data loading ticket goes in first...

mingjiecn (approved):

data/data_sources/data_sources_file_fileset.yaml
pedrohr: the only data loading ticket ready to QA is dual-ipa, it's already generated, so I think the JSONLs in this PR are the latest

PR #714, opened 2026-05-26
DSERV-1305-gwas

mingjiecn:


data/schemas/edges/variants_phenotypes.GWAS.json
```diff
@@ -88,9 +88,9 @@
           "description": "P value for the variant to phenotype edge",
           "example": 0.01
         },
-        "log10pvalue": {
+        "neg_log10_pvalue": {
           "type": "number",
-          "description": "Log10pvalue for the variant to phenotype edge",
+          "description": "neg_log10_pvalue for the variant to phenotype edge",
```
ian-whaling: "description": "Negative log10 of the p-value for the variant to phenotype edge",

src/routers/datatypeRouters/edges/variants_phenotypes.ts
```diff
@@ -286,7 +282,7 @@ async function findPhenotypesFromVariantSearch (input: paramsFormatType): Promis
 
         (record.source == 'OpenTargets' ? {
           study: ${input.verbose === 'true' ? `(${studyVerboseQuery})[0]` : 'record.study'},
-          ${getDBReturnStatements(variantPhenotypeGwasSchema, true, '', [], true, apiKeyToDbFieldMap)}
+          ${getDBReturnStatements(variantPhenotypeGwasSchema)}
```
ian-whaling, line 285: What is this change

mingjiecn: previously we need to map the database field to API field. Now the two fields are the same so we don't need apiKeyToDbFieldMap anymore. Other values are just default setting.

ian-whaling (approved):

PR #718, opened 2026-06-05
DSERV-1300 fixing E2G Crispr pvalue fields

pedrohr:


mingjiecn (changes requested):
The code use p_value = row[19] , which is wrong, row[18] should be p_value. Then you need to add p_value_adj using row[19]. Then calculate negative log10 for both of them and name them as neg_log10_p_value and neg_log10_p_value_adj

mingjiecn (approved):

mingjiecn (approved):

PR #719, opened 2026-06-08
DSERV-1312 Fixing source list and protein ID search

pedrohr:


mingjiecn (changes requested):
Need to sort collection_method here:
`collection_method = ', '.join(interaction_type)`

mingjiecn (approved):

PR #722, opened 2026-06-09
DSERV-1306 p-value changes for variants_proteins

pedrohr:


data/adapters/gvatdb_asb_adapter.py
```diff
@@ -84,15 +84,15 @@ def process_file(self):
                         '_from': _source,
                         '_to': _target,
                         'p_value': pvalue,
-                        'log10pvalue': log10pvalue,
+                        'neg_log10_pvalue': -1 * log10pvalue,
```
mingjiecn: I think log10pvalue is already negative value of log10pvalue( it should be positive in database) so you don't need to time -1 here. Can you double check?

data/adapters/pQTL_adapter.py
```diff
@@ -87,7 +87,7 @@ def process_file(self):
                             'label': 'pQTL',
                             'class': self.collection_class,
                             'method': self.method,
-                            'log10pvalue': float(row[14]),
+                            'neg_log10_pvalue': -1 * float(row[14]),
```
mingjiecn: same here. Can you double check the value?

data/tests/test_pQTL_adapter.py
```diff
@@ -33,7 +33,7 @@ def test_pQTL_adapter(mock_request, mocker):
         assert first_item['label'] == 'pQTL'
         assert first_item['method'] == 'pQTL'
         assert first_item['class'] == 'observed data'
-        assert first_item['log10pvalue'] == 79.2
+        assert first_item['neg_log10_pvalue'] == -79.2
```
mingjiecn: neg_log10_pvalue should be a positive number

mingjiecn (changes requested):

data/adapters/gvatdb_asb_adapter.py
pedrohr: oh you are right! just reverted the code

data/adapters/pQTL_adapter.py
pedrohr: just reverted the value

data/tests/test_pQTL_adapter.py
pedrohr: reverted

mingjiecn:
Can you also check the API to update it accordingly?

mingjiecn (approved):

mingjiecn (approved):

mingjiecn (changes requested):
Need to fix test

mingjiecn (approved):

PR #728, opened 2026-06-10
DSERV-1296-variant-painting

shengchd:


data/adapters/variant_painting_coding_variant_scores_adapter.py
```diff
@@ -0,0 +1,109 @@
...
+                    self.writer.write(json.dumps(_props))
+                    self.writer.write('\n')
+
+        if skipped_coding_variants:
+            with open(f'./skipped_coding_variants_{self.file_accession}.txt', 'a') as skipped_list:
+                for skipped in skipped_coding_variants:
+                    skipped_list.write(skipped + '\n')
+
+    def parse(self):
+        file_fileset_obj = get_file_fileset_by_accession_in_arangodb(
```
pedrohr, line 88: we need to add the S3 file tagging now. Example: https://github.com/IGVF-DACC/igvf-catalog/pull/800/changes#diff-352b00922c83dbef853e38eb431fc19f35c4af9cac4d581ba9cb03a94fe8bdfdR87-R92

shengchd: Oh I had it on line 93, will move it here to be consistent.

PR #729, opened 2026-06-10
DSERV-1242-more-perturb-seq-clean

ian-whaling:


src/routers/datatypeRouters/edges/genomic_elements_genes.ts
```diff
@@ -12,6 +12,10 @@ import { getSchema, getCollectionEnumValuesOrThrow } from '../schema'
 const MAX_PAGE_SIZE = 500
 const METHODS = getCollectionEnumValuesOrThrow('edges', 'genomic_elements_genes', 'method')
 const SOURCES = getCollectionEnumValuesOrThrow('edges', 'genomic_elements_genes', 'source')
+// genomic_elements_genes holds CRISPR E2G, Perturb-seq, ENCODE-rE2G, etc.
+// Resolve API p_value from whichever significance field the edge record stores.
+const RESOLVED_P_VALUE_AQL = '(record.neg_log10_p_value_adj != null ? record.neg_log10_p_value_adj : (record.neg_log10_p_value != null ? record.neg_log10_p_value : (record.log10pvalue != null ? record.log10pvalue : record.p_value)))'
```
mingjiecn: based on my understanding from last meeting. We will not use p_value to represent neg_log10_p_value_adj, neg_log10_p_value, log10pvalue(we should have no log10pvalue after reloading). Should return the original field name. Same score, should return original field name instead of score.

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -102,134 +784,166 @@ def process_file(self):
             self.file_accession)
         method = file_fileset['method']
         crispr_modality = file_fileset.get('crispr_modality')
+        targeted_element_types = self._targeted_element_types()
+        layout = self._layout()
+        layout_name = self._file_config().get('layout')
```
mingjiecn: some functions are called multiple times in the script, for example:
_file_config()
_layout()
Should just call once when initiating. Then use it directly:
self.file_config = self._file_config()

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -75,6 +79,684 @@ def _parse_bool(value):
...
+            'class': file_fileset['class'],
+            'name': 'modulates expression of',
+            'inverse_name': 'expression modulated by',
+            'method': method,
+            'crispr_modality': crispr_modality,
+            'biological_context': file_fileset['simple_sample_summaries'][0],
+            'biosample_term': file_fileset['samples'][0],
+            'treatments_term_ids': file_fileset['treatments_term_ids'],
+        }
+        if 'p_value' in metrics:
```
mingjiecn: can change to something like this:
```
OPTIONAL_FIELDS = {
    'pct_1',
    'pct_2',
    'effect_size',
    ...
}

for field in OPTIONAL_FIELDS:
    if field in metrics:
        edge[field] = metrics[field]
```

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -75,6 +79,684 @@ def _parse_bool(value):
...
+        candidate: dict,
+    ) -> bool:
+        if current is None:
+            return True
+        candidate_p = candidate.get('p_value')
+        current_p = current.get('p_value')
+        if candidate_p is None:
+            return False
+        if current_p is None:
+            return True
```
mingjiecn, line 714: why return true if current_p is None?

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -102,134 +784,166 @@ def process_file(self):
...
+                    readout_gene_raw = row[read_idx]
+                elif uses_name_hg38 or method == 'Perturb-seq':
+                    (
+                        intended_target_chr,
+                        intended_target_start,
+                        intended_target_end,
+                        intended_target_gene_raw,
+                    ) = self._resolve_perturb_seq_element(
+                        row, colmap, name_to_idx)
+                    read_idx = colmap['readout_gene']
```
mingjiecn: this line of code is duplicated with line 835: `read_idx = colmap['readout_gene']`
Should move it outside if else block so you only need to define it once.

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -75,6 +79,684 @@ def _parse_bool(value):
...
+        if not (c_raw and s_raw and e_raw):
+            return None
+        gene_raw = row[pi] if pi is not None and pi < len(row) else ''
+        return (c_raw, s_raw, e_raw, gene_raw)
+
+    def _resolve_name_hg38_interval(
+        self,
+        row: list,
+        col: Dict[str, Optional[int]],
+        name_to_idx: Optional[Dict[str, int]] = None,
```
mingjiecn: name_to_idx is not used

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -75,6 +79,684 @@ def _parse_bool(value):
...
+        pi = col['promoter_gene']
+        if pi is not None and pi < len(row) and row[pi].strip():
+            gene_raw = row[pi]
+        return chr_, start, end, gene_raw
+
+    def _resolve_perturb_seq_element(
+        self,
+        row: list,
+        col: Dict[str, Optional[int]],
+        name_to_idx: Optional[Dict[str, int]] = None,
```
mingjiecn: name_to_idx is actually not used

data/adapters/igvf_E2G_CRISPR_adapter.py
```diff
@@ -75,6 +79,684 @@ def _parse_bool(value):
...
+        seen_element_gene_ids: set,
+    ) -> None:
+        if element_gene_id in seen_element_gene_ids:
+            raise ValueError(
+                f'Duplicate element_gene edge in {self.file_accession}: {element_gene_id}'
+            )
+        seen_element_gene_ids.add(element_gene_id)
+
     def __init__(self, filepath, label, source_url, writer: Optional[Writer] = None, validate=False, **kwargs):
         self.source_url = source_url
```
mingjiecn, line 785: also we should be able to generate source_url from self.file_accession? We can remove source_url when initiate the adaper to make initiation simpler.

mingjiecn (changes requested):
Also Otto just updated base adapter. So you will need to rebase and adjust this adapter.

src/routers/datatypeRouters/edges/genomic_elements_genes.ts
ian-whaling: I started adjusting this but I think this ought to be resolved in a follow-up ticket to fix the API for genomic_elements_genes, coz it will probably need new indexes and harmonizing with the other E2G methods.

ian-whaling: I'm gonna revert all changes to API

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: Fixed.

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: fixed, thanks for the suggestion

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: avoids comparing None to a float and prefers a candidate with a p_value over a stored edge without one.
Added comment + test to make this explicit

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: fixed

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: fixed

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: fixed

data/adapters/igvf_E2G_CRISPR_adapter.py
ian-whaling: We could, but it is a current convention to require source_url, so I just also include it. If we seek to remove this redundancy should probably be a cross-codebase effort

mingjiecn (approved):

PR #732, opened 2026-06-15
DSERV-1330 using label for oncotree

pedrohr:


mingjiecn (changes requested):
Should not use type. Should use label to be consistent

data/adapters/oncotree_adapter.py
```diff
@@ -31,12 +31,12 @@ class Oncotree:
     API_URL = 'https://oncotree.mskcc.org:443/api/tumorTypes'
     SOURCE_URL = 'https://oncotree.mskcc.org/api/tumorTypes'
```
mingjiecn, line 34: Need to add constant `ALLOWED_LABELS` and validate the labels

mingjiecn (changes requested):

mingjiecn (approved):

PR #733, opened 2026-06-15
DSERV-1315 reading simple_sample_summaries from AnalysisSet obj

pedrohr:


ian-whaling (changes requested):
You will need to regenerate the jsonl for IGVF files_filesets in parsed_data

ian-whaling (approved):

PR #734, opened 2026-06-16
DSERV-1332-adapt-tagging

ottojolanki:


data/data_parser.py
```diff
@@ -22,6 +31,8 @@
                     help='The AWS profile to use, for example "igvf-dev".')
 parser.add_argument('--version-tag', type=str, default=None,
                     help='The version tag to use, for example "IGVF_catalog_beta_v0.4".')
+parser.add_argument('--other-tags', nargs='*', action=ParseOtherTagsAction,
```
mingjiecn: can you show an example here how to use this arg?

data/adapters/writer.py
```diff
@@ -48,12 +48,38 @@ def __init__(self, bucket: str, key: str, session: boto3.Session, version_tag: O
...
+                             for t in existing.get('TagSet', [])}
+        except client.exceptions.NoSuchKey:
+            existing_tags = {}
+
+        for tag in self.s3_tags:
+            key, value = tag['Key'], tag['Value']
+            if key in existing_tags:
+                current_values = existing_tags[key].split(' ')
+                if value not in current_values:
+                    current_values.append(value)
```
mingjiecn: why do we need to append old values?

mingjiecn (changes requested):

data/data_parser.py
ottojolanki:
```
--other-tags foo=bar baz='spam spam'

will create a tag `foo` with value `bar` and a tag `baz` with value `spam spam`

data/adapters/writer.py
ottojolanki: The idea is that if a tag has multiple values, they will be represented as space separated string, so for example:
before update the tag would be `portal_accessions: IGVFABC IGVFXYZ` and we add `IGVFTUV`.
In this case the current values will be `['IGVFABC', 'IGVFXYZ']`, we append the new value `IGVFTUV` to this list and create the new space separated tag value from the new list (in sorted order to be invariant on the order which the values are added).

mingjiecn (approved):

mingjiecn (approved):

PR #747, opened 2026-06-25
DSERV-1443-mutpred2-portal-accession-tags

ottojolanki:


mingjiecn (changes requested):
There is a mapping file `MAPPING_FILE` used in the adapter, need to add this file as well

mingjiecn (approved):

PR #752, opened 2026-06-26
DSERV-1445-mpra-portal-accession-tags

ottojolanki:


mingjiecn (changes requested):
Should we always update tag when initiating, making it easy to manage. In data_sources yaml file I do see encode file accession and encode source url. So if the accession here needs to be igvf, then data_sources yaml file needs to be updated

mingjiecn (approved):

pedrohr (approved):

PR #753, opened 2026-06-26
DSERV-1442-star-seq-portal-accession-tags

ottojolanki:


mingjiecn (approved):

data/adapters/STARR_seq_adapter.py
```diff
@@ -81,6 +81,7 @@ def parse(self):
         with open_file as f:
             reader = csv.reader(f, delimiter='\t')
             chunk = []
+            self.writer.add_tag('portal_accessions', self.file_accession)
```
ian-whaling: move this to line 78

ottojolanki: most likely this is a subtle timing bug related to the smart_open using a multipart upload. should be fixed in the writer class now.

PR #768, opened 2026-07-12
DSERV-1324-CRISPR-phenotype

ian-whaling:


data/index-config.yaml
```diff
@@ -310,6 +310,26 @@ genomic_elements_genes:
...
 
+genomic_elements_phenotypes:
+  persistent:
+    fields:
+      - method
+      - method, files_filesets
+      - source, files_filesets
+      - files_filesets
+      - files_filesets, _key
+      - z_score, files_filesets
```
pedrohr: we can drop the `z_score` since we don't have a query input for it in the API

data/index-config.yaml
```diff
@@ -310,6 +310,26 @@ genomic_elements_genes:
...
+      - _from, files_filesets
+      - _to, _key
+      - biosample_term, biological_context, source, method, files_filesets
+      - biological_context, source, method, files_filesets
+      - method, source, label, class, files_filesets
```
pedrohr: we can remove label and class, they are not in the API input

pedrohr (approved):
just need to remove a few indexes and it's fine

data/index-config.yaml
ian-whaling: done

data/index-config.yaml
ian-whaling: done

PR #771, opened 2026-07-14
DSERV-1473 adding files_filesets to GWAS

pedrohr:


data/adapters/gwas_adapter.py
```diff
@@ -189,9 +185,10 @@ def process_variants_phenotypes(self, row, tagged_variants):
             'version': 'October 2022 (22.10)',
             'name': 'associated with',
             'inverse_name': 'associated with',
-            'class': self.collection_class,
-            'method': self.method,
-            'label': self.method
+            'class': self.file_fileset.get('class'),
+            'method': self.file_fileset.get('method'),
+            'label': self.file_fileset.get('method'),
```
pedrohr, line 191: these might have to be changed according to QA, depending on whether files_filesets will have these attributes or will have to be hardcoded

data/adapters/gwas_adapter.py
```diff
@@ -28,19 +27,15 @@ class GWAS(BaseAdapter):
 
     MAX_LOG10_PVALUE = 27000  # max abs value on pval_exponent is 26677
     ONTOLOGY_MAPPING_PATH = './data_loading_support_files/gwas_ontology_term_name_mapping.pkl'
+    FILE_ACCESSION = 'IGVFFI1309WDQG'
```
mingjiecn: We should not hard code the FILE_ACCESSION and SOURCE_URL

mingjiecn (changes requested):

mingjiecn (approved):

PR #773, opened 2026-07-14
DSERV-1455-pharmGKB

mingjiecn:
updated the adapter to make sure we use the files from igvfd when parsing. We keep those files in data_loading_support_files folder but only use them for sample adapter.


data/data_sources/data_sources.yaml
```diff
@@ -1801,28 +1801,42 @@ drug:
...
   params:
     - label
+    - drug-reference-filepath
+    - variant-reference-filepath
+    - study-reference-filepath
   pypy3: false
-  command: python3 data_parser.py --adapter variant_drug --output-bucket igvf-catalog-parsed-collections --filepath {datafolder} --output-bucket-key variants_drugs/variants_drugs_pharmGKB.jsonl --label variant_drug
+  command: python3 data_parser.py --adapter variant_drug --output-bucket igvf-catalog-parsed-collections --filepath {datafolder} --drug-reference-filepath {drug_datafile} --variant-reference-filepath {variant_datafile} --study-reference-filepath {study_datafile} --output-bucket-key variants_drugs/variants_drugs_pharmGKB.jsonl --label variant_drug
   datafiles:
     - https://api.data.igvf.org/reference-files/IGVFFI8835SMSP/@@download/IGVFFI8835SMSP.tar.gz # variantAnnotations.tar.gz
```
shengchd, line 1816: should this file be pharmGKB_genes instead?

mingjiecn: > should this file be pharmGKB_genes instead?

pharmGKB_genes.tsv.gz is only used for variant_drug_gene

shengchd (approved):

PR #775, opened 2026-07-14
DSERV-1461 adding 3 new query parameters to variants/genes

pedrohr:


data/schemas/edges/variants_genes.AFGRSQtl.json
```diff
@@ -132,7 +144,7 @@
     }
   ],
   "accessible_via": {
-    "filter_by_range": "intron_start, intron_end, neg_log10_pvalue, effect_size",
+    "filter_by_range": "intron_start, intron_end, neg_log10_pvalue, effect_size, posterior_inclusion_probability, log2FC",
     "simplified_return": "class, method, label, source, neg_log10_pvalue, biological_context",
     "return": "intron_chr, intron_start, intron_end, effect_size, neg_log10_pvalue, source, class, method, label, p_value, chr, source_url, biological_context, biosample_term"
```
mingjiecn: should we also need to return those newly added fields?

mingjiecn (changes requested):

data/schemas/edges/variants_genes.AFGRSQtl.json
pedrohr: it's being returned already, but I can add to the official return list in the schema

mingjiecn (approved):

PR #780, opened 2026-07-16
DSERV-1485-crispr-variant-phenotype

ian-whaling:


data/adapters/CRISPR_variant_phenotype_adapter.py
```diff
@@ -0,0 +1,348 @@
...
+            'phenotype_name': 'LDL-C uptake',
+            'variant_id_col': 'variant_id',
+            'variant_type_col': 'target_group',
+            'variant_type_value': 'Variant',
+            'effect_size_col': 'mu_adj',
+            'z_score_col': 'mu_z_adj',
+            'num_guides_col': 'n_guides',
+            'edit_rate_mean_col': 'edit_rate_mean',
+            'ci_lower_col': 'CI[0.025',
+            'ci_upper_col': '0.975]',
```
pedrohr, line 62: are these supposed to be numbers?

pedrohr (changes requested):

data/adapters/CRISPR_variant_phenotype_adapter.py
ian-whaling: No that is how the field appears in the file
see https://data.igvf.org/tabular-files/IGVFFI6803HZJG/

ian-whaling: This mapping exists to map column variable to field name in file

pedrohr (approved):

PR #782, opened 2026-07-17
DSERV-1489-tar

mingjiecn:


data/adapters/archive_utils.py
```diff
@@ -0,0 +1,145 @@
...
+import tempfile
+import zipfile
+from pathlib import Path
+from typing import Iterator
+
+
+FILE_ACCESSION_PATTERN = re.compile(r'(?:IGVFFI|ENCFF)[A-Z0-9]+')
+
+
+def get_file_accession(filepath: str) -> str:
```
pedrohr, line 16: does this mean we need to use this helper in every adapter?

pedrohr (approved):

data/adapters/archive_utils.py
mingjiecn: We can use this function get_file_accession to get file accession so you don't need to write the code in the adapter.

PR #798, opened 2026-07-29
DSERV-1483-more-crispr-element-gene

ian-whaling:


ian-whaling:
Unrelated to the ticket I renamed all CRISPR adapters to have consistent naming

ian-whaling:
> I am fine with renaming the adapters, but that means fields `catalog_adapters ` in igvfd portal will need to be patched for all the files used by those adapters. Do you have the ticket created somewhere?

This field has largely not been kept in sync on the portal, so the effort to patch this property should be database-wide.

ian-whaling:
> > I am fine with renaming the adapters, but that means fields `catalog_adapters ` in igvfd portal will need to be patched for all the files used by those adapters. Do you have the ticket created somewhere?
> 
> This field has largely not been kept in sync on the portal, so the effort to patch this property should be database-wide.

https://igvf.atlassian.net/browse/WRN-1249

PR #800, opened 2026-07-29
DSERV-1475 adding files_filesets, class, and method to Hocomoco adapter

pedrohr:


data/schemas/edges/motifs_proteins.Motif.json
```diff
@@ -53,6 +53,32 @@
...
+      "type": "string",
+      "description": "The class of the motifs to proteins edge, from the associated files_filesets record",
+      "enum": [
+        "observed data"
+      ]
+    },
+    "method": {
+      "type": [
+        "string",
+        "null"
```
mingjiecn: can method be null for this source? If not, should remove it to make it strict.

data/schemas/nodes/motifs.Motif.json
```diff
@@ -19,11 +19,42 @@
...
+          "type": "string",
+          "description": "The class of the motif, from the associated files_filesets record",
+          "enum": [
+            "observed data"
+          ]
+        },
+        "method": {
+          "type": [
+            "string",
+            "null"
```
mingjiecn: remove null in method

